PR #1, opened 2023-02-21
Normalize

frostedoyster:
I'm finding that normalizing xyz before the calculation is more efficient than normalizing the spherical harmonics at the end


Luthaf:
This was implemented as part of #2

PR #2, opened 2023-02-22
Refactor the code to C++

Luthaf:
Still missing in this branch, and not part of the main branch

- [x] C API
- [x] Make Python package compatible with PEP 518 

To be imported from the main branch:

- [x] python module
- [x] docs
- [x] examples


ceriottm:
Does it make sense to start implementing here the normalization?
Also, to be discussed, do we want to provide a C++ class as well as the C API function calls?

Luthaf:
Added the C and Python API back! There seems to be an issue in the prefactor calculation, creating a `SphericalHarmonics` object twice gives a malloc error:

```
>>> import sphericart
>>> sphericart.SphericalHarmonics(20)
>>> sphericart.SphericalHarmonics(20)

python(97908,0x102cdc580) malloc: Incorrect checksum for freed object 0x137891000: probably modified after being freed.
Corrupt value: 0xbfa745d1745d1746
python(97908,0x102cdc580) malloc: *** set a breakpoint in malloc_error_break to debug
```

src/templates.hpp
```diff
@@ -176,14 +176,19 @@ void generic_sph(
...
+
+    #pragma omp parallel 
     {
         // thread-local storage arrays for Qlm (modified associated Legendre
         // polynomials) and terms corresponding to (scaled) cosine and sine of
         // the azimuth
-        double *q = (double *)malloc(sizeof(double) * (l_max + 1) * (l_max + 2) / 2);
-        double *c = (double *)malloc(sizeof(double) * (l_max + 1));
-        double *s = (double *)malloc(sizeof(double) * (l_max + 1));
+        double *q = new double[(l_max + 1) * (l_max + 2) / 2];
```
Luthaf, line 189: This could also be a `std::vector<double>`, removing  the need to remember to free/delete at the end.

ceriottm: I didn't know if it made anything less compatible. If not, happy to go with it!

Luthaf: I think if we assume C++17, we can use `std::vector` 😄 I'll do the change

PR #4, opened 2023-02-25
Add Torch bindings

Luthaf:
This only contains CPU bindings, and is restricted to float64, but should be a good starting point.

@nickjbrowning let me know if anything is unclear when you try to add a GPU version!

TODO:

- [x] add torch to wheels/sdist build
- [x] add torch to the documentation
- [x] add installation instructions


torch/core
```diff
@@ -0,0 +1 @@
+..
```
Luthaf, line 1: @ceriottm @frostedoyster @nickjbrowning bringing your attention to this. The `torch/core` folder is actually a symlink back to the root of the repo, allowing to include this code inside the sphericart_torch sdist. It might be a bit confusing though.

This is the cleanest way I found to be able to both build the torch bindings from the root project (for C++ workflows) and packaging the core source files in the source distribution of the sphericart_torch Python package, but if you have better ideas I'm listening!

ceriottm:
I can't build this in an environment that doesn't have cuda. I think we should allow for an installation of CPU-only sphericart-torch, although this may not be the most typical use case

Luthaf:
Yes, CI has the same issue. In the mean time, you can try with ˋpip install —no-build-isolation`, or using ˋpython setup.py bdist_wheel && pip install dist/sphericart_torch*.whl`

ceriottm:
I know this is still a WIP, but just a note that with the latest changes it does not even install the C version from the root folder. Might not be a bad idea to start and maintain some installation instructions in the `README`

PR #5, opened 2023-03-10
Refactor the C++ library to be more modular and better exploit templates

ceriottm:
This PR breaks does several things:
- Breaks down the monolithic generic_sph implementation into a bunch of inline functions that could - in principle - be called independently to compute a single sample, or a single l channel. 

- [x] TODO: Should we expose an API for these?

- Got rid of the Qlm temporary array. Qlms are computed on the fly - there is some duplication because the dYlm require Q(l-1)m, but overall this seems to be faster as well as more modular
-  All functions are now type-templated although it looks like single-precision makes execution *slower*


src/templates.hpp
```diff
@@ -15,7 +15,20 @@
 #include "macros.hpp"
 
 template <typename DTYPE, int HARDCODED_LMAX>
-inline void hardcoded_sph_template(DTYPE x, DTYPE y, DTYPE z, DTYPE x2, DTYPE y2, DTYPE z2, DTYPE *sph_i) {
+inline void hardcoded_sph_template(
+    DTYPE __attribute__((unused)) x, // these are unused in some code pathways, hopefully the compiler will compile them away
```
Luthaf, line 21: I think `__attribute__((unused))` is a GCC/Clang thing, not sure it would work with Intel compilers or MSVC.

ceriottm: went for `[[maybe_unused]]` which is in C++17 as `if constexpr`

src/sphericart.cpp
```diff
@@ -32,50 +33,63 @@ void sphericart::compute_sph_prefactors(int l_max, double *factors) {
...
-inline void _hardcoded_lmax_switch(int n_samples, int l_max, const double *xyz, double *sph, double *dsph) {
+// specialized versions
+void sphericart::compute_sph_prefactors(int l_max, double *factors) {
+    _compute_sph_prefactors<double>(l_max, factors);
+}
+
+// specialized versions
+void sphericart::compute_sph_prefactors(int l_max, float *factors) {
+    _compute_sph_prefactors<float>(l_max, factors);
+}
```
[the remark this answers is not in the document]
ceriottm: I agree with the idea and we could do this in the next refactoring iteration, when we also do a class implementation

Luthaf:
Ok, let's merge this and continue the refactor for the new API elsewhere!

PR #7, opened 2023-03-17
Class-based C++ API

ceriottm:
Use a SphericalHarmonics class to define the C++ interface.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -12,17 +12,20 @@ build:
     - cmake
   tools:
     python: "3.10"
-    rust: "1.61"
-
-# Build documentation in the docs/ directory with Sphinx
-sphinx:
-   configuration: docs/src/conf.py
+  jobs:
+    # install sphericart_torch with the CPU version of PyTorch
+    # we can not use the `python` section below since it does not allow
+    # to specify `--extra-index-url`
+    post_install:
+      - pip install --extra-index-url https://download.pytorch.org/whl/cpu ./torch
 
 # Declare the Python requirements required to build the docs
 python:
    install:
    - method: pip
      path: .
-   #- method: pip
-   #  path: ./torch
    - requirements: docs/requirements.txt
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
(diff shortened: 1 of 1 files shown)


Luthaf:
TODO before merging:

- [ ] check documentation

ceriottm:
> This is mostly fine for me.
> 
> Now that we hide the prefactors, we don't need to group everything in two objects, and we could have separate class members for different prefactors/buffers, which would increase readability (especially compared to the paper). But we can change this later!

Actually I tried this - I had even made a type holding `c,s,z` and used it for the buffer - but it was leading to noticeable slowdown of the calculation. Go figure.

frostedoyster:
Documentation (for reference): https://sphericart--7.org.readthedocs.build/en/7/

Luthaf:
From @ceriottm in https://github.com/lab-cosmo/scikit-matter/pull/171

> Only problem remaining: I noticed that the torch API documentation gets generated locally, but not on readthedocs. @Luthaf can you help? Then I'd merge so we can move to CUDA and start engaging external devs.

I'll give this a look tomorrow!

ceriottm:
Only problem remaining: I noticed that the torch API documentation gets generated locally, but not on readthedocs. @Luthaf can you help? Then I'd merge so we can move to CUDA and start engaging external devs.

ceriottm:
> From @ceriottm in [lab-cosmo/scikit-matter#171](https://github.com/lab-cosmo/scikit-matter/pull/171)
> 
> > Only problem remaining: I noticed that the torch API documentation gets generated locally, but not on readthedocs. @Luthaf can you help? Then I'd merge so we can move to CUDA and start engaging external devs.
> 
> I'll give this a look tomorrow!

Yeah sorry wrong repo ^_^

frostedoyster:
Generating documentation for torch requires the torch build to succeed on readthedocs, but it was broken so I disabled it. I assumed @Luthaf was working on it on the other branch. You'll need to uncomment the torch build in .readthedocs.yml and then fix the issues with the build

Luthaf:
Seems to be working with the CPU version of torch!

PR #8, opened 2023-03-27
Add CUDA version of the torch extension

Luthaf:


Luthaf:
This now compile for me both for CPU and CUDA version, but fails to run with ` RuntimeError: requested shared memory buffer (52976) exceeds max available on device: Quadro P1000 (49152)`.

PR #11, opened 2023-04-06
A function that replaces e3nn.o3.spherical_harmonics

ceriottm:
Idea is simple. While we wait for e3nn devs to decide if they want to directly integrate sphericart, we can just call `sphericart_torch.patch_e3nn(e3nn)` and substitute our function call for theirs.

This seems to work although I still need to figure out all combinations of normalization parameters.


torch/python/src/sphericart_torch/__init__.py
```diff
@@ -117,4 +125,70 @@ def compute(
...
+) -> torch.Tensor:
+    """
+    Computes spherical harmonics with an interface similar to the e3nn package.
+
+    Provides an interface that is similar to :py:func:`e3nn.o3.spherical_harmonics`
+    but uses :py:class:`SphericalHarmonics` for the actual calculation.
+    Uses the same ordering of the [x,y,z] axes, and supports the same options for
+    input and harmonics normalization as :py:mod:`e3nn`. However, it does not support
+    defining the irreps through a :py:class:`e3nn.o3._irreps.Irreps` or a string
+    specification, but just as a single integer or a list of integers.
```
Luthaf, line 142: How hard would it be to accept `Irreps` or string and turn them into a list of integers using code from e3nn?

torch/python/src/sphericart_torch/__init__.py
```diff
@@ -117,4 +125,70 @@ def compute(
...
+        shl = sh[:, l * l : (l + 1) * (l + 1)]
+        if normalization == "norm":
+            shl *= math.sqrt(1 / (2 * l + 1))
+        sh_list.append(shl)
+    sh = torch.cat(sh_list, dim=-1)
+
+    return sh
+
+
+def patch_e3nn(e3nn_module: ModuleType) -> None:
```
Luthaf, line 181: Should we also provide an `unpatch_e3nn`?

tox.ini
```diff
@@ -92,16 +91,14 @@ commands =
...
-    sphinx-build {posargs:-E} -W -b html docs/src docs/build/html
-
+passenv = 
+    PIP_EXTRA_INDEX_URL
 
-[testenv:fast-docs]
-# this environement re-builds the documentation without installing packages
 commands =
+    pip install .
+    pip install ./torch # --extra-index-url $PIP_EXTRA_INDEX_URL
```
Luthaf, line 99:
```suggestion
    pip install ./torch
```

torch/python/tests/test_e3nn.py
```diff
@@ -0,0 +1,71 @@
...
+# only run e3nn tests if e3nn is present
+try:
+    import e3nn
+
+    _HAS_E3NN = True
+except ModuleNotFoundError:
+    _HAS_E3NN = False
+
+torch.manual_seed(0)
+_PRECISION = 1e-8
```
Luthaf, line 18:
```suggestion
TOLERANCE = 1e-8
```

this is quite a high tolerance for float64. Do we really not do better than this?

torch/python/src/sphericart_torch/__init__.py
```diff
@@ -117,4 +125,70 @@ def compute(
...
+
+    if not hasattr(l_list, "__len__"):
+        l_list = [l_list]
+    l_max = max(l_list)
+    sh = SphericalHarmonics(l_max, normalized=normalize).compute(x[:, [2, 0, 1]])[0]
+
+    if normalization != "integral":
+        sh *= math.sqrt(4 * math.pi)
+
+    sh_list = []
```
Luthaf, line 170: Is there any performance difference with a list + cat vs torch.empty + assign?

torch/python/src/sphericart_torch/__init__.py
```diff
@@ -117,4 +125,70 @@ def compute(
...
+        a further scaling of the :math:`Y_m^l`. See the
+        documentation of :py:func:`e3nn.o3.spherical_harmonics()`
+        for a detailed explanation of the different conventions.
+    """
+
+    if not hasattr(l_list, "__len__"):
+        l_list = [l_list]
+    l_max = max(l_list)
+    sh = SphericalHarmonics(l_max, normalized=normalize).compute(x[:, [2, 0, 1]])[0]
+
```
Luthaf, line 166:
```suggestion
    assert normalization in ["integral", "norm", "component"]
```

torch/python/src/sphericart_torch/__init__.py
ceriottm: It's done in the spherical_harmonics implementation so we'd have to copy and paste a whole bunch of ifs, which I find a bit annoying.

```

        if isinstance(irreps_out, str):
            irreps_out = o3.Irreps(irreps_out)
        if isinstance(irreps_out, o3.Irreps) and irreps_in is None:
            for mul, (l, p) in irreps_out:
                if l % 2 == 1 and p == 1:
                    irreps_in = o3.Irreps("1e")
        if irreps_in is None:
            irreps_in = o3.Irreps("1o")

        irreps_in = o3.Irreps(irreps_in)
        if irreps_in not in (o3.Irreps("1x1o"), o3.Irreps("1x1e")):
            raise ValueError(f"irreps_in for SphericalHarmonics must be either a vector (`1x1o`) or a pseudovector (`1x1e`), not `{irreps_in}`")
        self.irreps_in = irreps_in
        input_p = irreps_in[0].ir.p  # pylint: disable=no-member

        if isinstance(irreps_out, o3.Irreps):
            ls = []
            for mul, (l, p) in irreps_out:
                if p != input_p**l:
                    raise ValueError(f"irreps_out `{irreps_out}` passed to SphericalHarmonics asked for an output of l = {l} with parity p = {p}, which is inconsistent with the input parity {input_p} — the output parity should have been p = {input_p**l}")
                ls.extend([l]*mul)
        elif isinstance(irreps_out, int):
            ls = [irreps_out]
        else:
            ls = list(irreps_out)
```

torch/python/src/sphericart_torch/__init__.py
ceriottm: True. Done.

torch/python/src/sphericart_torch/__init__.py
ceriottm: ok so I implemented the "do nothing if you don't need to do something" approach. should be good enough at least until we don't get specific requests from e3nn people.

torch/python/src/sphericart_torch/__init__.py
Luthaf: If we want this to be a drop-in replacement, I think we need to copy-paste this code. We should also patch the `SphericalHarmonics` torch module. 

This way people can install sphericart_torch, run this function and everything is faster even if people are not using e3nn directly (e.g. they are trying to use allegro)

ceriottm:
True, although painful...

frostedoyster (approved):

ceriottm:
Ok, good enough for the moment, and ended up mopping up a lot more fixes and tweaks. Merging now and moving on. Depending on how e3nn devs want to go about integration we may want to also patch the class version, and to handle the Irreps input

PR #15, opened 2023-04-19
Update Python build system

Luthaf:
This makes the sphericart-torch package install in such a way that we can import it with `import sphericart.torch`. One can also install both packages from the root with `pip install .[torch]`.

TODO:

- [x] update both README
- [x] update installation instructions


ceriottm:
so there was a bug that I had introduced when macro-ifying the CUDA code, I know one shouldn't mix PRs but I'm a bad boy, so I fixed it. sorry. otherwise, all works well. I've been able to compile & install on all the systems I have access to.

ceriottm:
I also added the documentation, I  think. @Luthaf pls check that the instructions from piPy are correct. Probably it'd be good to update the distribution soon after this is merged.

ceriottm:
Problem: making the C++ project with BUILD_TORCH enabled fails because folders have moved
```
 CMake Error at CMakeLists.txt:28 (add_subdirectory):
   add_subdirectory given source "torch" which is not an existing directory.
```

Luthaf:
Should be fixed now!

PR #16, opened 2023-04-20
Clean up the API and make the pytorch class torchscript compatible

ceriottm:
... this makes it less surprising that the two functions return different things.

Everything should be already done, including examples and docs, but another pair of eyes won't hurt before merging.


benchmarks/pytorch/benchmark.py
```diff
@@ -53,7 +53,7 @@ def sphericart_benchmark(
     time_noderi = np.zeros(n_tries + 10)
     for i in range(n_tries + 10):
         elapsed = -time.time()
-        sh_sphericart, _ = sh_calculator.compute(xyz, gradients=False)
+        sh_sphericart = sh_calculator.compute(xyz, gradients=False)
```
frostedoyster, line 56:
```suggestion
        sh_sphericart = sh_calculator.compute(xyz)
```

benchmarks/pytorch/benchmark.py
```diff
@@ -82,7 +82,7 @@ def sphericart_benchmark(
 
     for i in range(n_tries + 10):
         elapsed = -time.time()
-        sh_sphericart, _ = sh_calculator.compute(xyz, gradients=False)
+        sh_sphericart = sh_calculator.compute(xyz, gradients=False)
```
frostedoyster, line 85:
```suggestion
        sh_sphericart = sh_calculator.compute(xyz)
```

frostedoyster (approved):

benchmarks/pytorch/benchmark.py
ceriottm: I knew I had missed something!

python/src/sphericart/spherical_harmonics.py
```diff
@@ -84,23 +73,75 @@ def compute(
...
+            xyz_ptr = xyz.ctypes.data_as(ctypes.POINTER(ctypes.c_float))
+            sph_ptr = sph.ctypes.data_as(ctypes.POINTER(ctypes.c_float))
+
+            self._lib.sphericart_compute_array_f(
+                self._calculator_f, xyz_ptr, xyz_length, sph_ptr, sph_length
+            )
+
+        return sph
+
+    def compute_with_gradients(self, xyz: np.ndarray) -> Tuple[np.ndarray]:
```
Luthaf, line 94:
```suggestion
    def compute_with_gradients(self, xyz: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
```

the syntax for typing tuples is strange: Tuple[A] is a single element tuple, Tuple[A, B, A] is a three elements tuple, Tuple[A, ...] is a tuple of unknown length containing only A

examples/pytorch/torchscript.py
Luthaf, line 1: should this file be a test instead of an example? Or something similar made into a test? 

I don't think we are running the examples on CI (although maybe we should as well)

ceriottm: It was a leftover, I have integrated the example in the example, and added a separate test.
Running the examples in the CI might be a good idea given that I keep getting errors in those whenever we fiddle with the API.

ceriottm:
I think it's good to go. I also added the examples to the CI. this is a really big clean-up.

tox.ini
```diff
@@ -69,6 +70,25 @@ commands =
...
+deps =
+    wheel
+    cmake
+
+    numpy
+    torch
+    pytest
+
+passenv=
+    PIP_EXTRA_INDEX_URL
```
Luthaf, line 84: Is this explicitly required? I did not set it for other projects, and `PIP_EXTRA_INDEX_URL` seems to be passed through without any issue

ceriottm: IDK, it was not working but perhaps I made a mistake. Doesn't hurt, that for sure...

PR #17, opened 2023-04-21
Example TorchScript Compatibility Issue

nickjbrowning:
Added an example TorchScript file in examples/pytorch that shows the failure. We need to provide support for the class sphericart.torch.SphericalHarmonics to be TorchScript compatible.

### DIFF
--- a/examples/pytorch/torchscript.py
+++ b/examples/pytorch/torchscript.py
@@ -10,19 +10,9 @@ import torch
 docstring = """
 An example of the use of TorchScript and the PyTorch intrface of the `sphericart` library.
 
-Creates a TorchScript-compatible module to compute Cartesian spherical harmonics 
+Compiles the TorchScript-compatible module to compute Cartesian spherical harmonics 
 for an array of random 3D points. 
 """
-
-class SphericalHarmonicsModule(torch.nn.Module):
-    def __init__(self, lmax, normalized):
-        super().__init__()
-        self.sh_calculator = sphericart.torch.SphericalHarmonics(lmax, normalized=normalized)
-
-    def forward(self, xyz):
-        sph, _ = self.sh_calculator.compute(xyz)
-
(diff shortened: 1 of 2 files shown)


nickjbrowning:
Since we import and use torch already in the python wrapper here: https://github.com/lab-cosmo/sphericart/blob/main/sphericart-torch/python/sphericart/torch/__init__.py, we may just be able to directly convert the SphericalHarmonics class to a torch.nn.Module and this may fix the issue.

nickjbrowning:
My first idea failed - we need to add support directly to the C++ class that gets registered here:

```torch.classes.sphericart_torch.SphericalHarmonics```

PyTorch provides a nice example of how to do this here: https://pytorch.org/tutorials/advanced/torch_script_custom_classes.html

**Comment by frostedoyster** [comment]
I would really like the idea of converting our class to a `torch.nn.Module`. This is more idiomatic and it would allow people to send the class to and from GPU (in our case, this would move the prefactors, which should probably become a `torch.nn.Parameter`)

**Comment by nickjbrowning** [comment]
Torchscript doesn't like optional return types that change at run time it seems. I've modified the following code:

```python
def compute(
        self, xyz: torch.Tensor, gradients: bool = False
    ) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
        if gradients:
            return self._sph.compute_with_gradients(xyz)
        else:
            return self._sph.compute(xyz), None
```

to:

```python
def compute(
        self, xyz: torch.Tensor, gradients: bool = False
    ) -> List[torch.Tensor]:
        if gradients:
            return self._sph.compute_with_gradients(xyz)
        else:
            return self._sph.compute(xyz), torch.empty_like(xyz)
```

We'll need to adapt the docstring if we accept this suggested change....

ceriottm:
The example is now implemented in #16

PR #19, opened 2023-04-25
Verbose benchmarks

ceriottm:
More flexible handling of the "warmup phase" in pytorch tests, to also check for anomalous fluctuations in timings


ceriottm:
This is such a boring and painless PR that I'll just merge it

PR #21, opened 2023-04-25
Clean-up CUDA implementation

ceriottm:
An effort to clean-up and optimize the CUDA (and possibly CPU) implementation, reusing code whenever possible

A few action points, or at least points for discussion:

- [ ] more comments in the code, explaining the general strategy, and what each function does 
- [ ] removing all duplicate code re the C++ codebase
- [ ] if possible, make cuda.cu torch-agnostic which may help if we port to Jax

### DIFF
--- a/sphericart-torch/include/sphericart/torch.hpp
+++ b/sphericart-torch/include/sphericart/torch.hpp
@@ -61,7 +61,7 @@ private:
     torch::Tensor prefactors_cuda_float_;
 
     int64_t CUDA_GRID_DIM_X_ =  8;
-    int64_t CUDA_GRID_DIM_Y_ = 16;
+    int64_t CUDA_GRID_DIM_Y_ = 8;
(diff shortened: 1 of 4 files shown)


ceriottm:
> regarding the `generic_sph_l_channel_device`, there's a small change in the CUDA code. I pulled out the following to temporary variables as they are reused multiple times. I'm not sure if the compiler will simply optimise this away in the original code, but if I recall correctly I did notice a reduction in the timings when I did this.
> 
> ```c
>     scalar_t s_l = s[get_index(l)];
>     scalar_t s_l_neg1 = s[get_index(l - 1)];
>     scalar_t c_l = c[get_index(l)];
>     scalar_t c_l_neg1 = c[get_index(l - 1)];
> ```
> 
> In the for-loops I've done similar things to the above as well.

OK we can try this but I did test and on the CPU the compiler was smart enough to do the right thing, I would imagine the same may be true for the CUDA one. Worth a try.

PR #27, opened 2023-04-26
Windows support

Luthaf:
This is a WIP, trying to build the code and run the tests on Windows, using the MSVC compiler since that's the one Python requires.


sphericart/include/cuda_base.hpp
```diff
@@ -66,29 +66,29 @@ namespace cuda {
  */
 template <typename scalar_t>
 void spherical_harmonics_cuda_base(
-    const scalar_t* __restrict__ xyz,
+    const scalar_t* xyz,
     const int nedges,
-    const scalar_t* __restrict__ prefactors,
```
ceriottm, line 71: rather than removing __restrict__ maybe you can add an `#ifdef` clause that defines it to a noop when the compiler is not GCC

Luthaf:
Closing for now, we can re-open when we have time for it.

PR #31, opened 2023-04-27
illegal memory access fix

nickjbrowning:
fixed an issue with illegal memory accesses for small lmax.

Also significantly improves the small lmax timings as a consequence (but not higher lmax - this will be done in separate PR).

### DIFF
--- a/sphericart-torch/src/cuda.cu
+++ b/sphericart-torch/src/cuda.cu
@@ -262,7 +262,7 @@ __global__ void spherical_harmonics_kernel(
     int nl = max(
         static_cast<int>((HARDCODED_LMAX + 1) * (HARDCODED_LMAX + 1)),
          2 * lmax + 1
-     );
+    );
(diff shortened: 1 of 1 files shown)


sphericart-torch/src/cuda.cu
```diff
@@ -264,6 +264,8 @@ __global__ void spherical_harmonics_kernel(
          2 * lmax + 1
      );
 
+    int ml = min(static_cast<int>(HARDCODED_LMAX), lmax);
+
```
Luthaf, line 268: Could this be defined closer to where it is used? There are 100 lines between the variable definition and usage right now

ceriottm: I think the whole code needs a bit of internal documentation explaining what it is doing. Can I suggest we do this at a later stage? This is hardly the most confusing bit IMO ^_^

PR #32, opened 2023-04-27
Allow to disable -march=native from CMake

Luthaf:
Use this when building wheels on CI to make sure the wheels will run on any system


Luthaf:
From some quick testing, there is currently a 10% performance gain from using `-march=native`.

PR #33, opened 2023-04-27
improved global memory coalescing

nickjbrowning:
modified the CUDA code such that it uses additional threads whose only purpose (at the moment) is to write out the buffers to global memory with coalesced access.

### DIFF
--- a/sphericart-torch/src/autograd.cpp
+++ b/sphericart-torch/src/autograd.cpp
@@ -182,7 +182,7 @@ torch::autograd::variable_list SphericalHarmonicsAutograd::forward(
         dsph = results[1];
     } else if (xyz.device().is_cuda()) {
         // re-do the shared memory update in case `requires_grad` changed
-        const int GRID_DIM_Y = 1;
+        int GRID_DIM_Y = 16;
 
         const std::lock_guard<std::mutex> guard(calculator.cuda_shmem_mutex_);
 
@@ -205,7 +205,7 @@ torch::autograd::variable_list SphericalHarmonicsAutograd::forward(
             );
             printf("Re-attempting with GRID_DIM_X = 16\n");
 
-            calculator.CUDA_GRID_DIM_X_ = 16;
+            GRID_DIM_Y = 8;
             shm_result = calculator.cuda_shmem_.update_if_required(
                 xyz.scalar_type(),
                 calculator.l_max_,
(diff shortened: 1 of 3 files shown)


sphericart-torch/src/autograd.cpp
```diff
@@ -205,7 +205,7 @@ torch::autograd::variable_list SphericalHarmonicsAutograd::forward(
             );
             printf("Re-attempting with GRID_DIM_X = 16\n");
 
-            calculator.CUDA_GRID_DIM_X_ = 16;
+            GRID_DIM_Y = 8;
```
Luthaf, line 208: If we can change `GRID_DIM_Y`, we need to store it in the same way as we did GRID_DIM_X, to prevent spamming messages to the stdout

Luthaf: We should also change the message above

ceriottm:
I tested it on our Quadro RTX 6000, and I still get errors for l=0 (and actually also some other ls, like l=36 for some reason).  Error is triggered outside the cuda code but I suspect it's the usual CUDA weirdness 
```
Traceback (most recent call last):
  File "/home/ceriotti/source/sphericart/benchmarks/pytorch/benchmark.py", line 325, in <module>
    sphericart_benchmark(
  File "/home/ceriotti/source/sphericart/benchmarks/pytorch/benchmark.py", line 88, in sphericart_benchmark
    xyz = xyz.clone().detach().type(dtype).to(device).requires_grad_()
RuntimeError: CUDA error: an illegal memory access was encountered
```

ceriottm:
OK @nickjbrowning and I debugged this and I think it's good to go. Unless someone wants more time to review this, I'll merge later today.

sphericart-torch/src/cuda.cu
```diff
@@ -355,51 +356,55 @@ __global__ void spherical_harmonics_kernel(
...
+                    buffer_dsph_x,
+                    buffer_dsph_y,
+                    buffer_dsph_z,
+                    get_index
+                );
+            }
+        } else {
+            COMPUTE_SPH_L0(buffer_sph, get_index);
+	    if (requires_grad) {
+                COMPUTE_SPH_DERIVATIVE_L0(buffer_sph, buffer_dsph_x, buffer_dsph_y, buffer_dsph_z, get_index);
```
Luthaf, line 405:
```suggestion
            COMPUTE_SPH_L0(buffer_sph, get_index);
	        if (requires_grad) {
                COMPUTE_SPH_DERIVATIVE_L0(buffer_sph, buffer_dsph_x, buffer_dsph_y, buffer_dsph_z, get_index);
            }
```

Luthaf:
Looks fine to me. It would be nice to have a comment with a high-level explanation of what the different threads are doing.

PR #37, opened 2023-05-16
Remove unwanted }

kanduri:
Resolved build errors that were reported by Ilyes


kanduri:
This is the biggest reason why GPU CI coverage is super essential. PR coming for that as well shortly.

ceriottm:
Thanks. Indeed we really need to have this. Merging it immediately.

Issue #36, opened 2023-05-03
Add Jax bindings


ceriottm:
Tagging @felixmusil who seemed interested. Challenge here would be to reuse as much as possible the CUDA code @nickjbrowning wrote for libtorch. The whole concept is to keep the codebase as compact as possible.

felixmusil:
I looked into the details of how to extend JAX with custom [cpp and cuda](https://github.com/dfm/extending-jax) operations. 
From what I gather the heavy lifting consists in registering both the cpp and cuda code as `XLA custom` operations so that the operation can be jitted from the JAX side. On the cpp side, probably only some additional interface to follow the `XLA custom` op requirements is needed but the current cuda code depends on `torch` so I fear heavy refactoring would be required.

ceriottm:
I agree. I think that we need to start a discussion with @nickjbrowning in this respect.

nickjbrowning:
Hey @felixmusil !

I don't think it will be too tricky to separate out the CUDA code. As far as I can see we should be able to remove the torch dependencies and then provide two front-ends that wrap the PyTorch and XLA frontends separately. I'll be away next week but after that I can start looking into it when I'm back.

felixmusil:
Perfect then I will try to get the CPP side of the Jax bindings working in the meantime.

ceriottm:
I think the main issue will be packaging, but it's worth starting at least a draft PR

PR #42, opened 2023-05-22
JAX CPU interface

felixmusil:
Here is a draft for adding cpu bindings for the JAX framework discussed in #36. 

The current build is not integrated with a pip install so you need to follow [these instructions](https://github.com/lab-cosmo/sphericart/issues/36#issuecomment-1554803378) to test out the jax binding. Only the SPH+gradients have been binded so far. Will there be a need for a SPH only binding?

The main difference with the torch bindings is that JAX follows a functional paradigm so the precomputation of the SPH coefficients does not fall well with the current CPP implementation. So to simplify the PR, these coefficients are recomputed for each SPH computation call. 

I will try to follow how the torch bindings are setup for the packaging but I might need some help.

What kind of tests will be needed ? and what API should be provided to the users ?


felixmusil:
The JAX binding should now work with pip so `pip install .[jax]` should allow to do:

```
from jax.config import config
config.update('jax_platform_name', 'cpu')
config.update("jax_enable_x64", True)

import jax
import sphericart.jax as sph
key = jax.random.PRNGKey(0)
xyz = 6 * jax.random.normal(key,(4, 3))
l_max = 5
SPH = sph.spherical_harmonics(l_max,False,xyz)

def func(xyz):
    l_max = 2
    sph = spherical_harmonics(l_max, False, xyz)
    return sph.sum()
    
# can jit a function of the SPH
 jfunc = jit(func)
out = jfunc(xyz)
# take its gradient
dout = grad(jfunc)(xyz)

```

nickjbrowning:
this looks really promising Felix! I'll take a look over the coming days and see if we can unify the CUDA part into something jax-compatible in the coming days...

For unit tests, perhaps you could mirror the ones we've set up for sphericart-torch, located here:

https://github.com/lab-cosmo/sphericart/tree/main/sphericart-torch/python/tests

felixmusil:
Hey, I added the tests for the JAX bindings and cleaned a bit more my additions. So I think that it's now ready for review.

felixmusil:
I tried to adapt the current configuration to add `sphericart-jax` to the CI but it seems to fail at wheel building. I am not sure how to fix that since `pip install .[jax]` and `cd sphericart-jax/python/tests; pytest .` are working locally. Could you have a look?

ceriottm:
I tried to build locally, it complains about lack of pybind11

```
      CMake Error at CMakeLists.txt:10 (find_package):
        Could not find a package configuration file provided by "pybind11" with any
        of the following names:
      
          pybind11Config.cmake
          pybind11-config.cmake
      
        Add the installation prefix of "pybind11" to CMAKE_PREFIX_PATH or set
        "pybind11_DIR" to a directory containing one of the above files.  If
        "pybind11" provides a separate development package or SDK, be sure it has
        been installed.
```

felixmusil:
the last commit fixes your problem @ceriottm ? unfortunately it does not fix the CI...

sphericart-jax/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+        message(STATUS "Setting build type to 'relwithdebinfo' as none was specified.")
+        set(
+            CMAKE_BUILD_TYPE "relwithdebinfo"
+            CACHE STRING
+            "Choose the type of build, options are: none(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) debug release relwithdebinfo minsizerel."
+            FORCE
+        )
+        set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS release debug relwithdebinfo minsizerel none)
+    endif()
+endif()
```
Luthaf, line 33: This should be done directly after the `project()` call, otherwise cmake might need to run twice to configure everything right

sphericart-jax/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+    #  target_sources() converts relative paths to absolute
+    cmake_policy(SET CMP0076 NEW)
+endif()
+
+project(sphericart_jax CXX)
+
+find_package(pybind11 CONFIG REQUIRED)
+
+
+if(OpenMP_CXX_FOUND)
```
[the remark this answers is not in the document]
felixmusil: I have no idea, I mostly followed what's done in sphericart-torch. Do you mean that I should use `SPHERICART_OPENMP` from `sphericart/CMakeLists.txt` as a condition ?

sphericart-jax/CMakeLists.txt
felixmusil: done

sphericart-jax/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
+cmake_minimum_required(VERSION 3.10)
+
+if (POLICY CMP0076)
+    #  target_sources() converts relative paths to absolute
+    cmake_policy(SET CMP0076 NEW)
+endif()
+
+project(sphericart_jax CXX)
+
+find_package(pybind11 CONFIG REQUIRED)
```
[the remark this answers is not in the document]
felixmusil: this dependency is also taken care of when building with `pip install .[jax]`

sphericart-jax/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+            CMAKE_BUILD_TYPE "relwithdebinfo"
+            CACHE STRING
+            "Choose the type of build, options are: none(CMAKE_CXX_FLAGS or CMAKE_C_FLAGS used) debug release relwithdebinfo minsizerel."
+            FORCE
+        )
+        set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS release debug relwithdebinfo minsizerel none)
+    endif()
+endif()
+
+option(SPHERICART_JAX_BUILD_FOR_PYTHON "Are we building sphericart_jax for usage from Python?" OFF)
```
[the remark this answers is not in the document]
felixmusil: you are right there are no pure cpp build for the jax bindings.

sphericart-jax/CMakeLists.txt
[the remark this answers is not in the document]
Luthaf: No, I mean you should not have to use these flags, unless the jax-specific code uses `$pragma omp` somewhere. The core library is already compiled with OpenMP support

sphericart-jax/pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+repository = "https://github.com/lab-cosmo/sphericart"
+
+### ======================================================================== ###
+
+[build-system]
+requires = [
+    "setuptools >=44",
+    "wheel >=0.36",
+    "cmake",
+    "pybind11>=2.8.0",
```
Luthaf, line 46: Does this need to be `pybind11[global]`?

tox.ini
```diff
@@ -70,6 +71,26 @@ commands =
...
+
+    numpy
+    jax
+    pytest
+
+
+
+changedir = sphericart-jax
+passenv=
+    PIP_EXTRA_INDEX_URL
```
Luthaf, line 88: You should not need this, it is about allowing users to install a CPU-only torch package when running on Linux with a CUDA compiler.

sphericart-jax/CMakeLists.txt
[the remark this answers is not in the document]
Luthaf: I'm not sure then. If the installation of pybind11 with pip does not work, then you could try installing it with cmake, but I don't have a strong preference on either method.

Luthaf:
I don't have time to do a full review of this code, but maybe @frostedoyster can help? 

One thing to take into account is that this seems to be using a single function for the API, while all the other bindings are using a class with a `compute` member function.

felixmusil:
> One thing to take into account is that this seems to be using a single function for the API, while all the other bindings are using a class with a compute member function.

JAX is a functional programming framework so it seems more fitting to provide a function rather than a class. Moreover, the Cpp API does not allow to bind of a function to compute the coefficients and a function to compute the SPH with the coefficients as input so the resulting python class would only serve the purpose of following how it's done in the other bindings.

felixmusil:
I applied the changes after discussing the PR with @Luthaf in person. It should be good to go now!

sphericart-jax/python/sphericart/jax/sph_ops.py
```diff
@@ -0,0 +1,148 @@
...
+        set of n_sample vectors in 3D
+
+    Returns
+    -------
+    jax array [n_sample, (l_max+1)**2]
+        Spherical harmonics expension of `xyz`
+    """
+    ls = jnp.arange(l_max + 1)
+    output, _ = sph_fwd(ls, normalized, xyz)
+    return output
```
frostedoyster, line 145: It seems to me that we are evaluating both sph and dsph during the call to the spherical harmonics. That is however a large overhead if the user does not want to take derivatives. Would it be possible to calculate the derivatives only when people ask for computations that need the gradients?

sphericart-jax/python/sphericart/jax/sph_ops.py
```diff
@@ -0,0 +1,148 @@
...
+def sph_vjp(l_max, normalized, args, tangents):
+    dsph = args[0]
+    out = jnp.einsum("ndl,nl->nd", dsph, tangents)
+    return (out,)
+
+
+@partial(jax.custom_vjp, nondiff_argnums=(0, 1))
+def spherical_harmonics(l_max, normalized, xyz):
+    """Computes the Spherical harmonics and their derivatives within
+    the JAX framework. see `sphericart.SphericalHarmonics` for more details.
```
Luthaf, line 127:
```suggestion
    the JAX framework. See :py:class:`sphericart.SphericalHarmonics` for more details.
```

sphericart-jax/python/sphericart/jax/sph_ops.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Parameters
+    ----------
+    l_max : int
+        maximum order of the spherical harmonics (included)
+    normalized : bool
+        if the imput vectors have already been normalized
```
Luthaf, line 134:
```suggestion
        should we compute cartesian (``normalized=False``) or normalized spherical harmonics
```

sphericart-jax/python/sphericart/jax/sph_ops.py
felixmusil: There is little point of using JAX if one does not need gradients. The numpy implementation would be a better fit at that point.

frostedoyster: I agree that people use jax for gradients, but they don't have to be gradients of the spherical harmonics. For example, when you train a NN on energies, you need the gradients with respect to the weights, but not with respect to the spherical harmonics. Think also about higher derivatives: we won't be able to assume that users want the second derivatives at some point in their workflow... the only choice would be to calculate them on the fly

frostedoyster:
I'm not sure how it's working since I never define any vjp/transpose, but it works

frostedoyster:
Update: the jax `custom_call` interface changed slightly in the last few versions. We have to support the new one.
https://jax.readthedocs.io/en/latest/changelog.html
From jax 0.4.16:
A number of exports from jax.interpreters.xla related to defining HLO lowering rules for custom JAX primitives have been deprecated. Custom primitives should be defined using the StableHLO lowering utilities in jax.interpreters.mlir instead.

sphericart-jax/python/tests/test_autograd.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+import jax
+jax.config.update("jax_platform_name", "cpu")
+jax.config.update("jax_enable_x64", True)
```
Luthaf, line 4: This might fail on M1/ARM

sphericart-jax/python/tests/test_consistency.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@pytest.fixture
+def xyz():
+    key = jax.random.PRNGKey(0)
+    return 6 * jax.random.normal(key, (100, 3))
+
+
+def test_consistency(xyz):
+    xyz = jnp.array(xyz, dtype=jnp.float64)
```
Luthaf, line 17: why is this re-wraped in a `jnp.array`? Should already be one

sphericart-jax/python/tests/dummy_sph.py
```diff
@@ -0,0 +1,39 @@
+import jax
+import jax.numpy as jnp
+from functools import partial
+
+
+@partial(jax.vmap, in_axes=(0, None))
+def dummy_spherical_harmonics(xyz, l_max):
```
Luthaf, line 7:
```suggestion
def pure_jax_spherical_harmonics(xyz, l_max):
```

sphericart-jax/python/tests/test_dummy.py
Luthaf, line 1: file name suggestion: `test_transforms.py`

sphericart-jax/src/jax.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+template <typename T>
+void cpu_sph(void *out, const void **in) {
+    // std::cout << "Called kernel" << std::endl;
+    // Parse the inputs
+    const T *xyz = reinterpret_cast<const T*>(in[0]);
+    const size_t l_max = *reinterpret_cast<const int*>(in[1]);
+    const bool normalized = *reinterpret_cast<const bool*>(in[2]);
+    const size_t n_samples = *reinterpret_cast<const int*>(in[3]);
+    // std::cout << l_max << " " << normalized << " " << n_samples << std::endl;
```
Luthaf, line 23:
```suggestion
```

sphericart-jax/src/jax.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "sphericart.hpp"
+#include "sphericart/pybind11_kernel_helpers.h"
+
+using namespace sphericart_jax;
+
+namespace {
+
+template <typename T>
+void cpu_sph(void *out, const void **in) {
+    // std::cout << "Called kernel" << std::endl;
```
Luthaf, line 17:
```suggestion
```

sphericart-jax/src/jax.cpp
```diff
@@ -0,0 +1,90 @@
...
+    const size_t l_max = *reinterpret_cast<const int*>(in[1]);
+    const bool normalized = *reinterpret_cast<const bool*>(in[2]);
+    const size_t n_samples = *reinterpret_cast<const int*>(in[3]);
+    // std::cout << l_max << " " << normalized << " " << n_samples << std::endl;
+    size_t xyz_length{n_samples*3};
+    size_t sph_len{(l_max+1)*(l_max+1)*n_samples};
+
+    // The output is stored as a single pointer since there is only one output,
+    // hence no need to reinterpret_cast the out pointer
+    T *sph = reinterpret_cast<T *>(out);  // CHANGEEEEEEEEEEEEEEE SHOULD NOT NEED REINTERPRET_CAST
```
Luthaf, line 29: So what do we do here?

sphericart-jax/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+
+find_package(pybind11 CONFIG REQUIRED)
+
+
+
+set(BUILD_SHARED_LIBS OFF CACHE BOOL "" FORCE)
+add_subdirectory(sphericart EXCLUDE_FROM_ALL)
+
+
+# include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
```
Luthaf, line 32:
```suggestion
```

sphericart-jax/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+
+# CPU op library
+pybind11_add_module(cpu_ops ${CMAKE_CURRENT_LIST_DIR}/src/jax.cpp)
```
Luthaf, line 35:
```suggestion
pybind11_add_module(sphericart_jax ${CMAKE_CURRENT_LIST_DIR}/src/jax.cpp)
```

I think the library should be named sphericart_jax, especially if we install it in `<prefix>/lib`

sphericart-jax/python/sphericart/jax/utils.py
```diff
@@ -0,0 +1,2 @@
+def default_layouts(*shapes):
```
Luthaf, line 1: Do we really need a separate file just for this function?

sphericart-jax/python/sphericart/jax/spherical_harmonics.py
```diff
@@ -0,0 +1,28 @@
+from .sph import sph
+
+
+def spherical_harmonics(xyz, l_max, normalized):  # TODO: CHANGE xyz to first argument everywhere
```
Luthaf, line 4: What's the status on this TODO?

frostedoyster: It's done! Removed

sphericart-jax/src/jax.cpp
frostedoyster: I was wrong: it does need a cast from void* to T*

sphericart-jax/python/tests/test_autograd.py
frostedoyster: Ok! I'll take them out except for the test that checks precision, where it is necessary

sphericart-jax/python/tests/test_consistency.py
frostedoyster: True, it's gone

sphericart-jax/python/tests/dummy_sph.py
frostedoyster: Done!

sphericart-jax/CMakeLists.txt
frostedoyster: Changed to `sphericart_jax_cpu`

felixmusil (commented):
The PR looks great @frostedoyster. I am not sure why I couldn't make the forward and backward mode work before, amazing that you managed it.

frostedoyster:
Great, and thanks for getting this started @felixmusil! @Luthaf, could you have a final look?

Luthaf (approved):
I cleaned up some unused code in kernels_helpers.hpp, should be good to go!

PR #43, opened 2023-05-23
Enable GPU CI on Piz Daint

kanduri:
This currently builds only the PyTorch side of library runs only the benchmark script. We should also have unit tests in addition to end-to-end tests and benchmarks. 

If this works, I will also build the CPP API and test the benchmarks and examples there in the subsequent commits.


Luthaf:
I don't see a status check on this PR for the Daint CI, should we have one?

PR #53, opened 2023-06-13
Python CI testing using Tox

kanduri:


ceriottm:
Looks good. Anything else you want to add, or should we merge?

Luthaf:
I don't know when I'll have time to do this, so I'll put the idea here if someone else wants to give it a go in the mean time! 

The idea was to use the [`package=external` option](https://tox.wiki/en/latest/config.html#external-package-builder) in tox, which uses a separate environment to build the code and store a wheel, and then uses this pre-built wheel in all tests environments. We can only do this for one package, which would be sphericart-torch since it is the slowest to compile.

We are doing this in equistore for equistore-core, if you want to see an example of it.

kanduri:
That should be enough for me to take a stab at it.

PR #54, opened 2023-06-14
Improving the developer comments in the C++ implementation

ceriottm:
In this PR I try to address some concerns on the fact that the C++ core is kind of impenetrable. I added throughout comments explaining what is done and why. Is there anything else that would be useful to do while I've all the C++ files open?

### DIFF
--- a/sphericart/include/macros.hpp
+++ b/sphericart/include/macros.hpp
@@ -8,14 +8,20 @@
     values are computed for one point at a time, and that the spherical harmonics are
     stored in a contiguous section that "flattens" the (l,m) dimensions, e.g.
     [ (0,0), (1,-1), (1,0), (1,1), (2,-2), ...]
+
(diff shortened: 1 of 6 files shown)


Luthaf (commented):
This is a nice high level overview!

I'd also like to see some explanation what each function parameter & stack variable refers to in templates.hpp (ideally with reference to the paper/using the paper notation). For example what are `c`/`s`/`two_mz`/`rxy`/`pq`/`ql1m_2`/...

ceriottm:
> This is a nice high level overview!
> 
> I'd also like to see some explanation what each function parameter & stack variable refers to in templates.hpp (ideally with reference to the paper/using the paper notation). For example what are `c`/`s`/`two_mz`/`rxy`/`pq`/`ql1m_2`/..., and how are they indexed?

better?

Luthaf:
yes, this is much nicer!

PR #64, opened 2023-07-12
CUDA backward bug

frostedoyster:
At the moment, we decide whether to calculate double backward second derivatives by checking if `ddsph` is defined or not. This is fine for the CPU code, but on the CUDA code `dsph` and `ddsph` are defined also if these are not required. This leads to second derivatives of the wrong shape which cause torch to crash later on

### DIFF
--- a/sphericart-torch/python/tests/test_nn.py
+++ b/sphericart-torch/python/tests/test_nn.py
@@ -38,6 +38,12 @@ def test_nn():
     loss = (target - energy) ** 2 + torch.sum((d_target - forces) ** 2)
     loss.backward()
 
+    if torch.cuda.is_available():
+        nn = NN().to("cuda")
+        energy, forces = nn(xyz.detach().to("cuda"))
+        loss = (target.to("cuda") - energy) ** 2 + torch.sum((d_target.to("cuda") - forces) ** 2)
+        loss.backward()
+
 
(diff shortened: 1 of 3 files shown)


frostedoyster:
I don't know how to ping this, but the problematic parts are in the CUDA file, lines 723-751:
`    torch::Tensor d_sph;
    if (xyz.requires_grad() || gradients)
    {
        d_sph = torch::empty(
            {xyz.size(0), 3, n_total},
            torch::TensorOptions().dtype(xyz.dtype()).device(xyz.device()));
    }
    else
    {
        // just so accessor doesn't complain
        d_sph = torch::empty(
            {1, 1, 1},
            torch::TensorOptions().dtype(xyz.dtype()).device(xyz.device()));
    }

    torch::Tensor hess_sph;

    if (xyz.requires_grad() && hessian)
    {
        hess_sph = torch::empty(
            {xyz.size(0), 3, 3, n_total},
            torch::TensorOptions().dtype(xyz.dtype()).device(xyz.device()));
    }
    else
    {
        hess_sph = torch::empty(
            {1, 1, 1, 1},
            torch::TensorOptions().dtype(xyz.dtype()).device(xyz.device()));
    }`

Luthaf (approved):
The fix looks fine

PR #67, opened 2023-08-17
Fix large-size integer overflow

### LINKED ISSUE #41: Segmentation fault on large tensors
While computing spherical harmonics on large tensors, one may face the segmentation fault with signal 11. The following example should reproduce the error.

```
import torch
import sphericart.torch as sphericart

if __name__ == '__main__':
    calculator = sphericart.SphericalHarmonics(l_max=12, normalized=True)
    vectors = 5.0 * torch.rand(100000000, 3)
    sp_harm = calculator.compute(vectors)
```
---

Naive memory profiling of this script gives the following feedback:

```
/usr/bin/time -v python test_sphericart.py 
Command terminated by signal 11
        Command being timed: "python test_sphericart.py"
        User time (seconds): 3.54
        System time (seconds): 1.09
        Percent of CPU this job got: 64%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:07.22
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 2578860
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 67800
        Voluntary context switches: 8953
        Involuntary context switches: 97625
        Swaps: 0
        File system inputs: 0
        File system outputs: 0
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

frostedoyster:
Fixes #41.
@Luthaf is it possible to write a test for this? How much memory do we have on the test environments? Are any changes needed to the torch side?
@nickjbrowning how about CUDA?

### DIFF
--- a/sphericart/include/sphericart.hpp
+++ b/sphericart/include/sphericart.hpp
@@ -338,9 +338,9 @@ private:
     // function pointers are used to set up the right functions to be called
     // these are set in the constructor, so that the public compute functions can 
     // be redirected to the right implementation
-    void (*_array_no_derivatives)(const T*, T*, T*, T*, int, int, const T*, T*);
-    void (*_array_with_derivatives)(const T*, T*, T*, T*, int, int, const T*, T*);
-    void (*_array_with_hessians)(const T*, T*, T*, T*, int, int, const T*, T*);
+    void (*_array_no_derivatives)(const T*, T*, T*, T*, size_t, int, const T*, T*);
+    void (*_array_with_derivatives)(const T*, T*, T*, T*, size_t, int, const T*, T*);
+    void (*_array_with_hessians)(const T*, T*, T*, T*, size_t, int, const T*, T*);
     
     // these compute a single sample
     void (*_sample_no_derivatives)(const T*, T*, T*, T*, int, int, const T*, const T*, T*, T*, T*);
--- a/sphericart/include/templates.hpp
+++ b/sphericart/include/templates.hpp
@@ -237,7 +237,7 @@ void hardcoded_sph(
     T *sph,
     [[maybe_unused]] T *dsph,
     [[maybe_unused]] T *ddsph,
-    int n_samples,
+    size_t n_samples,
     [[maybe_unused]] int l_max_dummy = 0, // dummy variables to have a uniform interface with generic_sph
     [[maybe_unused]] const T *prefactors_dummy = nullptr,
     [[maybe_unused]] T *buffers_dummy = nullptr)
@@ -257,7 +257,7 @@ void hardcoded_sph(
         [[maybe_unused]] T *ddsph : pointer to the storage location for the second derivatives.
                         stored as for sph_i, with nine consecutive blocks associated to the nine possible
                         second derivative combinations
-        int n_samples: number of samples that have to be computed
+        size_t n_samples: number of samples that have to be computed
 
     */
     static_assert(!(DO_SECOND_DERIVATIVES && HARDCODED_LMAX > 1), "Hardcoded second derivatives are not implemented for l>1.");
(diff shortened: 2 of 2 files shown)


frostedoyster:
Performance to be tested, but I'm very optimistic because all the inner loops that are thread-specific still run on ints. (Only the sample dimension is changed to size_t, while the l, m dimensions should be way too small to cause issues: even with second derivatives and l=100 we should be talking about a size of 9*101^2 which is about 90000)

Luthaf:
> Performance to be tested, but I'm very optimistic because all the inner loops that are thread-specific still run on ints. 

On 64-bit CPU (where size_t is 64-bit), CPU operations on 32 or 64-bit *integers* should be roughly as fast. The only reason to use 32 bit would be to save on memory consumption or bandwidth, which is not the case here.


---

Writing a test is going to be hard, [Github Action runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) have around 10 GiB of RAM available.

Luthaf (commented):
The changes looks good. Do we need to also propagate the type changes to values computed using `n_samples`, or do we already use `auto` everywhere?

frostedoyster:
Ok, perfect. No, no other changes should be needed on the CPU side because we only shift pointers along the sample dimension and we never use `i_sample` or `n_sample` otherwise. Regarding the tests, I guess we could squeeze one in because the minimum size required to make it fail is 8.8GB (using floats), but it's probably not worth it.

I'll have a look at the CUDA code and run one test on CUDA before merging

frostedoyster:
Update: the GPU implementation breaks

```
Traceback (most recent call last):
  File "break-sphericart.py", line 13, in <module>
    sp_harm = calculator.compute(vectors)
RuntimeError: numel needs to be smaller than int32_t max; otherwise, please use packed_accessor64
```

frostedoyster:
@nickjbrowning

nickjbrowning:
yeah changing this to packed_accessor64 should resolve it on the GPU side.

nickjbrowning:
the above commit changes the accessors to 64bit and the atomic indices to `size_t`.

I've also used this commit to piggyback a modified the `setup.py` slightly to resolve a potential build issue depending on how the users install the CUDA libraries (and math libraries). My change makes it work for the Nvidia HPC SDK.  Essentially the code conditionally adds an extra argument to `cmake_options`  that sets `DCUDA_TOOLKIT_ROOT_DIR` if the environment variable `CUDA_HOME` is set.

sphericart-torch/setup.py
```diff
@@ -21,12 +21,16 @@ def run(self):
...
+        CUDA_HOME = os.environ.get("CUDA_HOME")
+
         cmake_options = [
             f"-DCMAKE_INSTALL_PREFIX={install_dir}",
             "-DSPHERICART_TORCH_BUILD_FOR_PYTHON=ON",
             f"-DPYTHON_EXECUTABLE={sys.executable}",
             f"-DSPHERICART_ARCH_NATIVE={SPHERICART_ARCH_NATIVE}",
         ]
+        if CUDA_HOME is not None:
+            cmake_options.append(f"-DCUDA_TOOLKIT_ROOT_DIR={CUDA_HOME}")
```
Luthaf, line 33:
```suggestion
        cmake_options = [
            f"-DCMAKE_INSTALL_PREFIX={install_dir}",
            "-DSPHERICART_TORCH_BUILD_FOR_PYTHON=ON",
            f"-DPYTHON_EXECUTABLE={sys.executable}",
            f"-DSPHERICART_ARCH_NATIVE={SPHERICART_ARCH_NATIVE}",
        ]
        
        CUDA_HOME = os.environ.get("CUDA_HOME")
        if CUDA_HOME is not None:
            cmake_options.append(f"-DCUDA_TOOLKIT_ROOT_DIR={CUDA_HOME}")
```

Luthaf (approved):
Looks good, I would just move the CUDA_HOME search closer to where it is used in setup.py

PR #68, opened 2023-09-04
Enforce C/C++ formatting

frostedoyster:


frostedoyster:
I can't tell why the documentation is failing

Luthaf:
> ModuleNotFoundError: No module named 'torch._C._VariableFunctions'; 'torch._C' is not a package


It looks like the torch installation is broken??

.clang-format
```diff
@@ -0,0 +1,137 @@
...
+    SortPriority:    0
+  - Regex:           '.*'
+    Priority:        1
+    SortPriority:    0
+IncludeIsMainRegex: '(Test)?$'
+IncludeIsMainSourceRegex: ''
+IndentCaseLabels: false
+IndentGotoLabels: true
+IndentPPDirectives: None
+IndentWidth:     2
```
Luthaf, line 85: if you are fine with it, I'd prefer an indent width of 4, I find it makes it easier to see the different nesting levels. If you have strong feelings, I can live with these settings!

frostedoyster: Sure, I'll change it

frostedoyster:
Of course it works locally

Luthaf:
Are you running locally with the same packages versions as CI? You can remove & recreate the `.tox` directory, and you should get the latest version of everything, but otherwise tox prints the exact installed versions in CI log.

frostedoyster:
Yes, I deleted `.tox/` and it still runs fine

PR #71, opened 2023-10-20
Fix sphericart-jax bugs when calling vmap of derivatives

frostedoyster:


ceriottm (approved):
Looks great and fixes the problem.

sphericart-jax/python/tests/test_transforms.py
```diff
@@ -64,3 +64,23 @@ def test_hessian_jit(xyz):
...
+            
+            
+def test_vmap_grad(xyz):
+    def single_scalar_output(x, l_max, normalized):        
+        return jax.numpy.sum(sphericart.jax.spherical_harmonics(x, l_max, normalized))
+    single_grad = jax.grad(single_scalar_output)
+    sh_grad = jax.vmap(single_grad, in_axes=(0, None, None), out_axes=0)
+    for l_max in [4, 7, 10]:
+        for normalized in [True, False]:
+            _ = sh_grad(xyz, l_max, normalized)
```
Luthaf, line 76: shouldn't this have an assert somewhere?

PR #72, opened 2023-10-21
Improve JAX benchmark

frostedoyster:


ceriottm:
Looks good - but I get some strange timings on e3nn with small lmax and I wonder if there's something fishy. with 
```
python benchmark.py -l 5 -s 100 -t 1000 --warmup 5 --verbose --compare
```
I get lower time computing gradients rather than direct... 
```
 E3NN no der (jit):         712.8 ns/sample ±       58.8 (std)
 E3NN grad (jit):          477.2 ns/sample ±       62.5 (std)
```

PR #77, opened 2023-10-27
Updated CI to use an image with Python 3.10 installed

kanduri:
Once this is merged with main, any branch that builds on top of main should be able to trigger the CI with a new image with Python 3.10 in the environment


frostedoyster (approved):
Thank you very much @kanduri!!

PR #80, opened 2023-10-28
Disentangle hardcoding and templating

frostedoyster:
Following the discussion in #79, I realized that the templating of `l_max` (which leads to loop unrolling) and hardcoding can have different `l_max` thresholds. This would let us reap the benefits of loop unrolling also for `l_max>6`, although I haven't done any speed tests for the moment. An additional benefit is that this would remove the ugly duplication of code in `templates.hpp`, where we can now have a single loop which is fully static. (Here I'm assuming that our new `SPERICART_LMAX_TEMPLATED` would be so high that at that point loop unrolling wouldn't be possible anymore, so that we wouldn't realistically need to try enforcing unrolling for the first part of super-high `l`s and the compiler will choose for us at what point it stops unrolling.)
The only issue is template instantiation. I'm doing it with if statements at the moment (and `SPERICART_LMAX_TEMPLATED=10`) to avoid having to instantiate all the other template parameters, which would be a nightmare. Let me know what you think @ceriottm @Luthaf.

### DIFF
--- a/sphericart/include/templates.hpp
+++ b/sphericart/include/templates.hpp
@@ -19,8 +19,8 @@
 #define SPHERICART_LMAX_TEMPLATED 10
 
 #include <cmath>
-#include <vector>
 #include <iostream>
+#include <vector>
 
 #ifdef _OPENMP
 
@@ -639,7 +639,8 @@ CUDA_DEVICE_PREFIX static inline void generic_sph_l_channel(
 }
 
 template <typename T, bool DO_DERIVATIVES, bool DO_SECOND_DERIVATIVES,
-          int HARDCODED_LMAX, int TEMPLATED_L, int (*GET_INDEX)(int) = dummy_idx>
+          int HARDCODED_LMAX, int TEMPLATED_L,
+          int (*GET_INDEX)(int) = dummy_idx>
 CUDA_DEVICE_PREFIX static inline void generic_sph_l_channel_templated(
     [[maybe_unused]] T x, // these might be unused for low LMAX. not worth a
                           // full separate implementation
@@ -653,11 +654,10 @@ CUDA_DEVICE_PREFIX static inline void generic_sph_l_channel_templated(
     [[maybe_unused]] T *dzdy_sph_i, [[maybe_unused]] T *dzdz_sph_i) {
 
     generic_sph_l_channel<T, DO_DERIVATIVES, DO_SECOND_DERIVATIVES,
-                        HARDCODED_LMAX>(
-    TEMPLATED_L, x, y, z, rxy, pk, qlmk, c, s, twomz, sph_i, dx_sph_i, dy_sph_i,
-    dz_sph_i, dxdx_sph_i, dxdy_sph_i, dxdz_sph_i, dydx_sph_i,
-    dydy_sph_i, dydz_sph_i, dzdx_sph_i, dzdy_sph_i, dzdz_sph_i);
-
+                          HARDCODED_LMAX>(
+        TEMPLATED_L, x, y, z, rxy, pk, qlmk, c, s, twomz, sph_i, dx_sph_i,
+        dy_sph_i, dz_sph_i, dxdx_sph_i, dxdy_sph_i, dxdz_sph_i, dydx_sph_i,
+        dydy_sph_i, dydz_sph_i, dzdx_sph_i, dzdy_sph_i, dzdz_sph_i);
 }
 
 template <typename T, bool DO_DERIVATIVES, bool DO_SECOND_DERIVATIVES,
@@ -766,7 +766,8 @@ generic_sph_sample(const T *xyz_i, T *sph_i, [[maybe_unused]] T *dsph_i,
     // also initialize the sine and cosine, even if these never change
     c[0] = 1.0;
     s[0] = 0.0;
-    for (m = 1; m < TEMPLATED_LMAX + 1; ++m) {  // allow unrolling of the static part of the loop
+    for (m = 1; m < TEMPLATED_LMAX + 1;
+         ++m) { // allow unrolling of the static part of the loop
         c[m] = c[m - 1] * x - s[m - 1] * y;
         s[m] = c[m - 1] * y + s[m - 1] * x;
         twomz[m] = twomz[m - 1] + twoz;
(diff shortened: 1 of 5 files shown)


sphericart/include/templates.hpp
```diff
@@ -14,7 +14,12 @@
     different numbers of terms computed with hard-coded expressions.
 */
 
+// Defines the largest l_max for which we have templated implementations,
+// which can benefit from loop unrolling
+#define SPHERICART_LMAX_TEMPLATED 10
+
 #include <cmath>
+#include <iostream>
```
Luthaf, line 22:
```suggestion
```

Luthaf:
What do the benchmarks look like before/after this change?

cortner:
Note that the Julia "templated" code does more than just unroll. The key point is that it is able to keep all work on the stack, wheres a "flexible" code would need to use memory from the heap. This is one of the more problematic performance pitfalls in Julia.

frostedoyster:
I wasn't seeing any performance increase until I used this compiler-specific keyword (`__attribute__((always_inline))`). What I think is going on is that here
https://github.com/lab-cosmo/sphericart/blob/4295669a9a885eadd2e5df4dc4fafb2a6c134893/sphericart/include/templates.hpp#L641-L661
we have a TEMPLATED_L that is being fed as a function parameter to an "inline" function. The issue is that `inline` is just a suggestion to the compiler, and it can be ignored. If it is ignored, then TEMPLATED_L is not known and the loops inside the second function won't be unrolled.

frostedoyster:
```
****************** DOUBLE PRECISION ******************
Call without derivatives (no hardcoding) took 98.8981 ± 0.605256 ns / sample
Call with derivatives (no hardcoding) took 337.11 ± 1.67443 ns / sample
Call with second derivatives (no hardcoding) took 1336.61 ± 2.80391 ns / sample

Sample without derivatives took 106.918 ± 1.36062 ns / sample
Sample with derivatives took 255.564 ± 6.15337 ns / sample
Sample with second derivatives took 853.075 ± 86.4209 ns / sample

Call without derivatives took 78.1945 ± 0.471599 ns / sample
Call with derivatives took 302.576 ± 1.69244 ns / sample
Call with second derivatives took 1197.27 ± 8.23424 ns / sample
Call without derivatives (normalized) took 87.2571 ± 0.521217 ns / sample
Call with derivatives (normalized) took 353.931 ± 7.2684 ns / sample
Call with second derivatives (normalized) took 2393.58 ± 3.87189 ns / sample

================ Low-l timings ===========
L=1 (no h-c) values              took 4.71103 ± 0.164448 ns / sample
L=1 (no h-c) values+derivatives  took 6.90779 ± 0.129 ns / sample
L=1 hardcoded values             took 1.60676 ± 0.0933306 ns / sample
L=1 hardcoded values+derivatives took 3.61443 ± 0.117717 ns / sample

L=2 (no h-c) values              took 8.635 ± 0.178832 ns / sample
L=2 (no h-c) values+derivatives  took 8.63524 ± 0.142311 ns / sample
L=2 hardcoded values             took 3.41483 ± 0.114826 ns / sample
L=2 hardcoded values+derivatives took 13.1237 ± 0.225533 ns / sample

L=3 (no h-c) values              took 12.7715 ± 0.283376 ns / sample
L=3 (no h-c) values+derivatives  took 27.6311 ± 0.305504 ns / sample
L=3 hardcoded values             took 11.9532 ± 0.204903 ns / sample
L=3 hardcoded values+derivatives took 21.4224 ± 0.232066 ns / sample

L=4 (no h-c) values              took 19.1465 ± 0.265671 ns / sample
L=4 (no h-c) values+derivatives  took 46.6761 ± 0.323387 ns / sample
L=4 hardcoded values             took 9.95259 ± 0.155389 ns / sample
L=4 hardcoded values+derivatives took 37.4713 ± 0.329868 ns / sample

L=5 (no h-c) values              took 22.1213 ± 0.256414 ns / sample
L=5 (no h-c) values+derivatives  took 70.7344 ± 0.658015 ns / sample
L=5 hardcoded values             took 15.3568 ± 0.2742 ns / sample
L=5 hardcoded values+derivatives took 56.6069 ± 0.547587 ns / sample

L=6 (no h-c) values              took 31.4157 ± 0.364123 ns / sample
L=6 (no h-c) values+derivatives  took 105.335 ± 1.11829 ns / sample
L=6 hardcoded values             took 22.308 ± 0.304872 ns / sample
L=6 hardcoded values+derivatives took 87.8669 ± 1.19881 ns / sample
```

frostedoyster:
And this is the main branch:
```
****************** DOUBLE PRECISION ******************
Call without derivatives (no hardcoding) took 104.425 ± 0.57164 ns / sample
Call with derivatives (no hardcoding) took 300.287 ± 1.87295 ns / sample
Call with second derivatives (no hardcoding) took 1342.6 ± 2.97078 ns / sample

Sample without derivatives took 109.738 ± 2.0463 ns / sample
Sample with derivatives took 266.269 ± 95.4806 ns / sample
Sample with second derivatives took 853.931 ± 77.1591 ns / sample

Call without derivatives took 80.3673 ± 0.451071 ns / sample
Call with derivatives took 330.482 ± 1.50251 ns / sample
Call with second derivatives took 1357.74 ± 13.4713 ns / sample
Call without derivatives (normalized) took 89.9167 ± 0.639336 ns / sample
Call with derivatives (normalized) took 379.961 ± 3.39652 ns / sample
Call with second derivatives (normalized) took 2438.88 ± 3.19248 ns / sample

================ Low-l timings ===========
L=1 (no h-c) values              took 3.76709 ± 0.116769 ns / sample
L=1 (no h-c) values+derivatives  took 8.30298 ± 0.169697 ns / sample
L=1 hardcoded values             took 1.61501 ± 0.321693 ns / sample
L=1 hardcoded values+derivatives took 3.60945 ± 0.156557 ns / sample

L=2 (no h-c) values              took 7.19116 ± 0.112577 ns / sample
L=2 (no h-c) values+derivatives  took 7.19082 ± 0.122308 ns / sample
L=2 hardcoded values             took 3.43054 ± 0.10358 ns / sample
L=2 hardcoded values+derivatives took 13.1471 ± 0.18961 ns / sample

L=3 (no h-c) values              took 11.9002 ± 0.190871 ns / sample
L=3 (no h-c) values+derivatives  took 29.3307 ± 0.287947 ns / sample
L=3 hardcoded values             took 11.9445 ± 0.189964 ns / sample
L=3 hardcoded values+derivatives took 21.4568 ± 0.212694 ns / sample

L=4 (no h-c) values              took 18.6302 ± 0.208289 ns / sample
L=4 (no h-c) values+derivatives  took 47.7342 ± 0.364463 ns / sample
L=4 hardcoded values             took 9.95611 ± 0.134674 ns / sample
L=4 hardcoded values+derivatives took 38.019 ± 0.309817 ns / sample

L=5 (no h-c) values              took 27.5422 ± 0.323073 ns / sample
L=5 (no h-c) values+derivatives  took 69.7555 ± 0.565318 ns / sample
L=5 hardcoded values             took 15.2972 ± 0.1696 ns / sample
L=5 hardcoded values+derivatives took 56.8899 ± 0.594101 ns / sample

L=6 (no h-c) values              took 39.1725 ± 0.361854 ns / sample
L=6 (no h-c) values+derivatives  took 107.807 ± 1.31283 ns / sample
L=6 hardcoded values             took 22.193 ± 0.341033 ns / sample
L=6 hardcoded values+derivatives took 88.1554 ± 1.20667 ns / sample
```

Luthaf:
It does not look like a big improvement on the benchmarks. Some cases are faster, other slower but not by much more than 5%.

> The issue is that inline is just a suggestion to the compiler, and it can be ignored.

Yes, `inline` does *NOT* mean inline this function, and never had. Instead it says that the code for this function is provided inline (typically in a header), and one of the copies (whichever one) should be kept in the final binary. This is related to One Definition Rule (ODR).

If we want/need to use `__attribute__((always_inline))`, we'll have to check that the compiler supports it first in cmake. This is to ensure we can also support compilers that are not GCC (MSVC on Windows, the Intel compilers and the IBM compilers). We could also see if there is a standard C++17 attribute (looking like this `[[<some_name>]]`) instead.

frostedoyster:
You're right: the performance improvement doesn't seem to be there. I had a look and `__attribute__((always_inline))` is very specific to gcc

PR #83, opened 2023-11-06
Cuda separated

nickjbrowning:
WIP for separating out the CUDA implementation from pytorch. A total rewrite of the hackathon contribution.


sphericart-torch/CMakeLists.txt
```diff
@@ -112,18 +112,25 @@ find_package(Torch 1.13 REQUIRED)
...
 else()
     target_sources(sphericart_torch PUBLIC "src/cuda_stub.cpp")
 endif()
 
+if (CMAKE_CUDA_COMPILER)
+    target_sources(sphericart_torch PUBLIC "src/torch_cuda_wrapper.cpp")
+else()
+    target_sources(sphericart_torch PUBLIC "src/torch_cuda_wrapper_stub.cpp")
+endif()
+
```
Luthaf, line 133: I would merge the two if here!

sphericart-torch/include/sphericart/cuda.hpp
```diff
@@ -1,26 +1,26 @@
...
 
-#include <ATen/Tensor.h>
-
-#include <vector>
-
 namespace sphericart_torch {
 
-bool adjust_cuda_shared_memory(at::ScalarType scalar_type, int64_t l_max,
+template <typename scalar_t>
+void spherical_harmonics_cuda_base(
```
Luthaf, line 7: Is this the main function for users who want to use sphericart CUDA without going through torch?

If so, should we move it out of `sphericart-torch` and into `sphericart`?

Luthaf: This function needs documentation

sphericart-torch/include/sphericart/cuda.hpp
```diff
@@ -1,26 +1,26 @@
...
+    const scalar_t *__restrict__ xyz, const int nedges,
+    const scalar_t *__restrict__ prefactors, const int nprefactors,
+    const int64_t l_max, const bool normalize, const int64_t GRID_DIM_X,
+    const int64_t GRID_DIM_Y, const bool xyz_requires_grad,
+    const bool gradients, const bool hessian, scalar_t *__restrict__ sph,
+    scalar_t *__restrict__ dsph, scalar_t *__restrict__ ddsph);
+
+template <typename scalar_t>
+void spherical_harmonics_backward_cuda_base(
```
Luthaf, line 16: this function needs documentation

sphericart-torch/include/sphericart/cuda.hpp
```diff
@@ -1,26 +1,26 @@
...
+    const scalar_t *__restrict__ dsph, const scalar_t *__restrict__ sph_grad,
+    const int nedges, const int ntotal, scalar_t *__restrict__ xyz_grad);
+
+bool adjust_cuda_shared_memory(size_t element_size, int64_t l_max,
```
Luthaf, line 20: this function needs documentation

sphericart-torch/include/sphericart/torch.hpp
```diff
@@ -18,7 +18,7 @@ class CudaSharedMemorySettings {
         : scalar_size_(0), l_max_(-1), grid_dim_x_(-1), grid_dim_y_(-1),
           requires_grad_(false), requires_hessian_(false) {}
 
-    bool update_if_required(torch::ScalarType scalar_type, int64_t l_max,
+    bool update_if_required(size_t scalar_size, int64_t l_max,
```
Luthaf, line 21: this function needs documentation

Luthaf (commented):
I can't comment much on the CUDA code itself. The function organization is a bit weird to me though. I think the main point is on the API itself: should the provide the same class-based interface to the code for CUDA or should we use this function-based interface? If possible, I would prefer to provide users with the same API for the CPU & CUDA versions of sphericart.

----

It would also be nice to have an example of how to use this code without torch!

sphericart-torch/include/sphericart/cuda.hpp
nickjbrowning: yeah this is the main function. We should move cuda_base.cu to sphericart and have only the torch wrapper (torch_cuda_wrapper.cpp) present in sphericart-torch.

The documentation for spherical_harmoncis_cuda_base is already in the .cu file:

/*
    Wrapper to launch the CUDA kernel. Returns a vector containing the spherical
   harmonics and their gradients if required, otherwise returns the spherical
   harmonics and an empty tensor.

    GRID_DIM_X is the number of threads to launch in the x dimension. Used to
   parallelize over the sample dimension. GRID_DIM_Y is the number of threads to
   launch in the y dimension. Used only to improve memory throughput on reads
   and writes.

    Total number of threads used is GRID_DIM_X * GRID_DIM_Y.
*/

sphericart-torch/CMakeLists.txt
nickjbrowning: I think it's easier for me to provide a CUDA C example of how to use these functions when we've completed relocating the CUDA code into `sphericart`.

Traditionally CUDA C is function-based. If you want to make an additional wrapper around these we'd need to write it...

Luthaf: A C example is fine with me as well!

sphericart-torch/include/sphericart/cuda.hpp
Luthaf: ok, we should move the documentation to the header then, since that's where users typically expect to find it. The snippet above should also be expanded with a description of each parameter.

sphericart-torch/include/sphericart/cuda.hpp
nickjbrowning: Added documentation in latest commit 9da4642.

sphericart-torch/include/sphericart/cuda.hpp
nickjbrowning: Added documentation in latest commit https://github.com/lab-cosmo/sphericart/commit/9da4642c88c808b6b5e38478e7899f269bc49f03.

sphericart-torch/include/sphericart/torch.hpp
nickjbrowning: Added documentation in latest commit https://github.com/lab-cosmo/sphericart/commit/9da4642c88c808b6b5e38478e7899f269bc49f03.

sphericart-torch/include/sphericart/cuda.hpp
nickjbrowning: Fixed this in the latest commit.

sphericart-torch/CMakeLists.txt
```diff
@@ -112,16 +112,20 @@ find_package(Torch 1.13 REQUIRED)
...
     "src/autograd.cpp"
     "src/torch.cpp"
 )
 
-if (CMAKE_CUDA_COMPILER)
-    target_sources(sphericart_torch PUBLIC "src/cuda.cu")
+if (CMAKE_CUDA_COMPILER AND SPHERICART_ENABLE_CUDA)
+#if (CMAKE_CUDA_COMPILER)
+    #target_sources(sphericart_torch PUBLIC "src/cuda_base.cu")
+    target_sources(sphericart_torch PUBLIC "src/torch_cuda_wrapper.cpp")
```
Luthaf, line 125:
```suggestion
    target_sources(sphericart_torch PUBLIC "src/torch_cuda_wrapper.cpp")
```

sphericart-torch/CMakeLists.txt
```diff
@@ -112,16 +112,20 @@ find_package(Torch 1.13 REQUIRED)
...
 else()
-    target_sources(sphericart_torch PUBLIC "src/cuda_stub.cpp")
+    #target_sources(sphericart_torch PUBLIC "src/cuda_stub.cpp")
+    target_sources(sphericart_torch PUBLIC "src/torch_cuda_wrapper_stub.cpp")
```
Luthaf, line 128:
```suggestion
    target_sources(sphericart_torch PUBLIC "src/torch_cuda_wrapper_stub.cpp")
```

sphericart-torch/src/torch_cuda_wrapper.cpp
```diff
@@ -0,0 +1,157 @@
...
+            d_sph.data_ptr<double>(), hess_sph.data_ptr<double>());
+        break;
+    case torch::ScalarType::Float:
+        sphericart::cuda::spherical_harmonics_cuda_base<float>(
+            xyz.data_ptr<float>(), xyz.size(0), prefactors.data_ptr<float>(),
+            prefactors.size(0), l_max, normalize, GRID_DIM_X, GRID_DIM_Y,
+            xyz.requires_grad(), gradients, hessian, sph.data_ptr<float>(),
+            d_sph.data_ptr<float>(), hess_sph.data_ptr<float>());
+        break;
+    }
```
Luthaf, line 89: Should this use `AT_DISPATCH_FLOATING_TYPES`? It looks like a manual re-implementation of this macro.

sphericart-torch/src/torch_cuda_wrapper.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+    cudaDeviceSynchronize();
+
+    if (!gradients)
+        d_sph = torch::Tensor();
+    if (!hessian)
+        hess_sph = torch::Tensor();
+    return {sph, d_sph, hess_sph};
```
Luthaf, line 97:
```suggestion
    if (!gradients) {
        d_sph = torch::Tensor();
    }
    
    if (!hessian) {
        hess_sph = torch::Tensor();
    }
    
    return {sph, d_sph, hess_sph};
```

Let's avoid `goto fail`-style errors https://www.synopsys.com/blogs/software-security/understanding-apple-goto-fail-vulnerability-2.html

sphericart-torch/src/torch_cuda_wrapper.cpp
```diff
@@ -0,0 +1,157 @@
...
+        throw std::runtime_error(
+            "internal error: CUDA version called on non-CUDA tensor");
+    }
+
+    auto xyz_grad = torch::Tensor();
+
+    if (xyz.requires_grad()) {
+        xyz_grad = torch::empty_like(xyz);
+
+        switch (xyz.scalar_type()) {
```
Luthaf, line 116: same with `AT_DISPATCH_FLOATING_TYPES`

sphericart-torch/src/torch_cuda_wrapper.cpp
```diff
@@ -0,0 +1,157 @@
...
+torch::Tensor sphericart_torch::prefactors_cuda(int64_t l_max,
+                                                at::ScalarType dtype) {
+    auto result =
+        torch::empty({(l_max + 1) * (l_max + 2)},
+                     torch::TensorOptions().device("cpu").dtype(dtype));
+
+    if (dtype == c10::kDouble) {
+        compute_sph_prefactors(l_max, static_cast<double *>(result.data_ptr()));
+    } else if (dtype == c10::kFloat) {
+        compute_sph_prefactors(l_max, static_cast<float *>(result.data_ptr()));
```
Luthaf, line 150: I think `data_ptr` has a template parameter defaulting to `void`.


```suggestion
        compute_sph_prefactors(l_max, result.data_ptr<double>());
    } else if (dtype == c10::kFloat) {
        compute_sph_prefactors(l_max, result.data_ptr<float>());
```

sphericart-torch/src/torch_cuda_wrapper_stub.cpp
```diff
@@ -0,0 +1,34 @@
...
+    Torch wrapper for the CUDA kernel forwards pass.
+*/
+std::vector<torch::Tensor> sphericart_torch::spherical_harmonics_cuda(
+    torch::Tensor xyz, torch::Tensor prefactors, int64_t l_max, bool normalize,
+    int64_t GRID_DIM_X, int64_t GRID_DIM_Y, bool gradients, bool hessian) {
+
+    throw std::runtime_error(
+        "sphericart_torch was not compiled with CUDA support");
+
+    return {torch::Tensor()};
```
Luthaf, line 15:
```suggestion
```

return is not required if we unconditionally throw

sphericart-torch/src/torch_cuda_wrapper_stub.cpp
```diff
@@ -0,0 +1,34 @@
...
+/*
+    Torch wrapper for the CUDA kernel backwards pass.
+*/
+torch::Tensor sphericart_torch::spherical_harmonics_backward_cuda(
+    torch::Tensor xyz, torch::Tensor dsph, torch::Tensor sph_grad) {
+
+    throw std::runtime_error(
+        "sphericart_torch was not compiled with CUDA support");
+
+    return {torch::Tensor()};
```
Luthaf, line 27:
```suggestion
```

sphericart/CMakeLists.txt
```diff
@@ -13,7 +13,9 @@ option(BUILD_SHARED_LIBS "Build shared libraries instead of static ones" OFF)
 OPTION(SPHERICART_BUILD_TESTS "Build and run tests for Sphericart" OFF)
 OPTION(SPHERICART_OPENMP "Try to use OpenMP when compiling Sphericart" ON)
 OPTION(SPHERICART_ARCH_NATIVE "Try to use -march=native when compiling Sphericart" ON)
+option(SPHERICART_ENABLE_CUDA "Are we building the CUDA backend of Sphericart?" ON)
 
+set(CMAKE_CUDA_ARCHITECTURES native)
```
Luthaf, line 18: I'm a bit worried about setting this as the unconditional default, since it will produce binaries that only run on one system. Maybe we can re-use the same option as for CPU native arch?

```suggestion
if (SPHERICART_ARCH_NATIVE AND "${CMAKE_CUDA_ARCHITECTURES}" STREQUAL "")
    set(CMAKE_CUDA_ARCHITECTURES native)
endif()
```

sphericart/CMakeLists.txt
```diff
@@ -30,13 +32,44 @@ if (${CMAKE_CURRENT_SOURCE_DIR} STREQUAL ${CMAKE_SOURCE_DIR})
     endif()
 endif()
 
-add_library(sphericart
+set(COMMON_SOURCES
```
Luthaf, line 35: Instead of adding to the sources, I think you can keep the library declaration and add to it with `target_sources`, like we do for sphericart-torch.

sphericart/CMakeLists.txt
```diff
@@ -30,13 +32,44 @@ if (${CMAKE_CURRENT_SOURCE_DIR} STREQUAL ${CMAKE_SOURCE_DIR})
...
+#    "include/sphericart.hpp"
+#    "include/sphericart.h"
+#)
+
+# Find CUDA
+include(CheckLanguage)
+check_language(CUDA)
+if(CMAKE_CUDA_COMPILER AND SPHERICART_ENABLE_CUDA)
+    enable_language(CUDA)
+    set(CUDA_USE_STATIC_CUDA_RUNTIME OFF CACHE BOOL "" FORCE)
```
Luthaf, line 54: Why is this required?

sphericart/CMakeLists.txt
```diff
@@ -30,13 +32,44 @@ if (${CMAKE_CURRENT_SOURCE_DIR} STREQUAL ${CMAKE_SOURCE_DIR})
...
+endif()
+
+# Append the relevant CUDA files to sources
+list(APPEND COMMON_SOURCES "include/cuda.hpp")
+if (CMAKE_CUDA_COMPILER AND SPHERICART_ENABLE_CUDA)
+    list(APPEND COMMON_SOURCES "src/cuda_base.cu")
+else()
+    list(APPEND COMMON_SOURCES "src/cuda_stub.cpp")
+endif()
+message("Value of SPHERICART_ENABLE_CUDA: ${SPHERICART_ENABLE_CUDA}")
```
Luthaf, line 69:
```suggestion
```

sphericart/src/cuda_base.cu
```diff
@@ -148,17 +137,41 @@ __device__ inline void write_buffers(
...
+                ddsph[edge_idx][0][1][offset + i] = tmp_dxdy;
+                ddsph[edge_idx][0][2][offset + i] = tmp_dxdz;
+
+                ddsph[edge_idx][1][0][offset + i] = tmp_dydx;
+                ddsph[edge_idx][1][1][offset + i] = tmp_dydy;
+                ddsph[edge_idx][1][2][offset + i] = tmp_dydz;
+
+                ddsph[edge_idx][2][0][offset + i] = tmp_dzdx;
+                ddsph[edge_idx][2][1][offset + i] = tmp_dzdy;
+                ddsph[edge_idx][2][2][offset + i] = tmp_dzdz;*/
```
Luthaf, line 174:
```suggestion
```

sphericart/src/cuda_base.cu
```diff
@@ -88,14 +78,13 @@ __device__ inline void write_buffers(
...
-    bool requires_grad, bool requires_hessian, bool normalize) {
-    if (atom_idx < natoms) {
+    scalar_t *buffer_dsph_dzdz, scalar_t *sph, scalar_t *dsph, scalar_t *ddsph,
+    size_t n_total, bool requires_grad, bool requires_hessian, bool normalize) {
+    if (edge_idx < nedges) {
         for (int i = threadIdx.x; i < n_elements; i += blockDim.x) {
-            sph[atom_idx][offset + i] = buffer_sph[get_index(i)];
+
+            sph[edge_idx * n_total + offset + i] = buffer_sph[get_index(i)];
+            // sph[edge_idx][offset + i] = buffer_sph[get_index(i)];
```
Luthaf, line 87:
```suggestion
```

sphericart/src/cuda_base.cu
```diff
@@ -148,17 +137,41 @@ __device__ inline void write_buffers(
...
-
-                ddsph[atom_idx][1][0][offset + i] = tmp_dydx;
-                ddsph[atom_idx][1][1][offset + i] = tmp_dydy;
-                ddsph[atom_idx][1][2][offset + i] = tmp_dydz;
-
-                ddsph[atom_idx][2][0][offset + i] = tmp_dzdx;
-                ddsph[atom_idx][2][1][offset + i] = tmp_dzdy;
-                ddsph[atom_idx][2][2][offset + i] = tmp_dzdz;
+                // ddsph: nedges, 3, 3, nl -> [nedges * 9 * nl + i * 3 * nl + j
+                // * nl + k]
```
Luthaf, line 141:
```suggestion
```

sphericart/src/cuda_base.cu
```diff
@@ -175,9 +188,16 @@ __device__ inline void write_buffers(
                     tmp_dz = (tmp_dz - z * tmp) * ir;
                 }
 
-                dsph[atom_idx][0][offset + i] = tmp_dx;
-                dsph[atom_idx][1][offset + i] = tmp_dy;
-                dsph[atom_idx][2][offset + i] = tmp_dz;
+                /*dsph[edge_idx][0][offset + i] = tmp_dx;
+                dsph[edge_idx][1][offset + i] = tmp_dy;
+                dsph[edge_idx][2][offset + i] = tmp_dz;*/
```
Luthaf, line 193:
```suggestion
```

examples/cuda/example.cu
```diff
@@ -0,0 +1,82 @@
...
+
+    // initializes samples
+    auto xyz = std::vector<double>(n_samples * 3, 0.0);
+    for (size_t i = 0; i < n_samples * 3; ++i) {
+        xyz[i] = (double)rand() / (double)RAND_MAX * 2.0 - 1.0;
+    }
+
+    // to avoid unnecessary allocations, calculators can use pre-allocated
+    // memory, however one can provide uninitialized vectors that will be
+    // automatically reshaped
```
Luthaf, line 45:
```suggestion
    // memory, one also can provide uninitialized vectors that will be
    // automatically reshaped
```

examples/cuda/example.cu
```diff
@@ -0,0 +1,82 @@
...
+        std::vector<double>(n_samples * 3 * 3 * (l_max + 1) * (l_max + 1), 0.0);
+
+    /* ===== API calls ===== */
+
+    // internal buffers and numerical factors are initalized at construction
+    sphericart::cuda::SphericalHarmonics<double> calculator_cuda(l_max);
+    // note: the following line
+    // auto calculator_cuda =
+    // sphericart::cuda::SphericalHarmonics<double>(l_max); calls the copy
+    // constructor, which has been deleted, in some nvcc versions
```
Luthaf, line 59: why is it deleted only for some nvcc versions? Also, should we provide a move constructor to handle this?

sphericart-torch/src/torch.cpp
```diff
@@ -4,17 +4,23 @@
...
       calculator_double_(l_max_, normalized_),
       calculator_float_(l_max_, normalized_),
-      prefactors_cuda_double_(prefactors_cuda(l_max, c10::kDouble)),
-      prefactors_cuda_float_(prefactors_cuda(l_max, c10::kFloat)) {
+
+      calculator_cuda_double_(l_max_, normalized_),
+      calculator_cuda_float_(l_max_, normalized_) //,
+
+// prefactors_cuda_double_(prefactors_cuda(l_max, c10::kDouble)),
+// prefactors_cuda_float_(prefactors_cuda(l_max, c10::kFloat))
```
Luthaf, line 22:
```suggestion
```

sphericart-torch/src/torch_cuda_wrapper_stub.cpp
Luthaf: Still valid

sphericart/CMakeLists.txt
Luthaf: still valid

sphericart/CMakeLists.txt
Luthaf: still valid

sphericart/CMakeLists.txt
```diff
@@ -30,13 +31,49 @@ if (${CMAKE_CURRENT_SOURCE_DIR} STREQUAL ${CMAKE_SOURCE_DIR})
...
+endif()
+
+add_library(sphericart ${COMMON_SOURCES})
+
+if (CMAKE_CUDA_COMPILER AND SPHERICART_ENABLE_CUDA AND SPHERICART_ARCH_NATIVE)
+    set_target_properties(sphericart PROPERTIES CUDA_ARCHITECTURES native)
+endif()
+
+#set_property(TARGET sphericart PROPERTY CUDA_ARCHITECTURES native)
+#set_target_properties(sphericart PROPERTIES CUDA_ARCHITECTURES "80")
```
Luthaf, line 75:
```suggestion
```

sphericart/include/sphericart_cuda.hpp
```diff
@@ -0,0 +1,121 @@
...
+     *  @param normalized
+     *      If `false` (default) computes the scaled spherical harmonics, which
+     * are homogeneous polynomials in the Cartesian coordinates of the input
+     * points. If `true`, computes the normalized spherical harmonics that are
+     * evaluated on the unit sphere. In practice, this simply computes the
+     * scaled harmonics at the normalized coordinates \f$(x/r, y/r, z/r)\f$, and
+     * adapts the derivatives accordingly.
+     */
+    SphericalHarmonics(size_t l_max, bool normalized = false);
+
```
Luthaf, line 78: I think we need to explicitly define the move and copy constructors here. Move constructor should be fine, the copy one should create a new mutex.


```suggestion
    SphericalHarmonics(const SphericalHarmonics& other);
    SphericalHarmonics(SphericalHarmonics&& other);
    
    SphericalHarmonics& operator=(const SphericalHarmonics& other);
    SphericalHarmonics& operator=(SphericalHarmonics&& other);
```

sphericart/src/sphericart_cuda.cu
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    const std::lock_guard<std::mutex> guard(this->cuda_shmem_mutex_);
+
+    bool shm_result = this->cuda_shmem_.update_if_required(
+        sizeof(T), this->l_max, this->CUDA_GRID_DIM_X_, this->CUDA_GRID_DIM_Y_,
+        compute_with_gradients, compute_with_hessian);
+
+    if (!shm_result) {
+        printf("Warning: Failed to update shared memory specification with");
```
Luthaf, line 117: This should use `std::cerr` to print to the standard error (or not print at all unless some environment variable is set to prevent standard stream pollution)

examples/cuda/example.cu
nickjbrowning: I don't remember writing this and I don't think it's true either?

sphericart/include/sphericart_cuda.hpp
nickjbrowning: we'd also need to add these to the main `sphericart.hpp` class as well then.

frostedoyster:
We still need to include the CUDA example in the documentation. I'd be happy to do it

frostedoyster (approved):
Good to go for me. However, we should open an issue trying to finalize and harmonize the APIs between CUDA, C++/Python and Julia. Until we do that, I don't see the point of documenting the CUDA C++ API. All the rest is there IMO

PR #87, opened 2023-11-24
Fix Thread-safety in julia code

cortner:
When the harmonics were evaluated in a multi-threaded code this led to different threads writing to the same temporary arrays. This PR fixes that. It simply requires replacing the object pool with a thread-safe version.


cortner:
I'm confused why the PR thinks there are more than 3 commits to merge. I'll have to figure out what happened. If anybody can suggest what I did wrong, I'd be grateful.

cortner:
I better close this, create a new fork, rebase the changes, and make a new PR. sorry for the noise.

PR #88, opened 2023-11-24
Fix Julia code thread safety

cortner:
(Second attempt)

When the harmonics were evaluated in a multi-threaded code this led to different threads writing to the same temporary arrays. This PR fixes that. It simply requires replacing the object pool with a thread-safe version.


Luthaf (approved):
This looks fine! `TSafe` is also exported by `ObjectPools`?

PR #95, opened 2024-01-16
Cudastream implementation for non-default cuda streams.

nickjbrowning:


sphericart-torch/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ check_language(CUDA)
 if(CMAKE_CUDA_COMPILER)
     enable_language(CUDA)
     set(CUDA_USE_STATIC_CUDA_RUNTIME OFF CACHE BOOL "" FORCE)
+    add_compile_definitions(CUDA_AVAILABLE)
```
Luthaf, line 22: this should be 
```suggestion
    target_compile_definitions(sphericart_torch PRIVATE CUDA_AVAILABLE)
```
or 
```suggestion
    target_compile_definitions(sphericart_torch PUBLIC CUDA_AVAILABLE)
```

`PRIVATE` if this is only set when compiling sphericart, `PUBLIC` if this is to be defined when using pre-compiled sphericart in another project.

Luthaf (commented):
looks good overall, just a comment on CMake code!

sphericart-torch/CMakeLists.txt
nickjbrowning: I've committed the PRIVATE statement.

Luthaf (approved):
I'll merge once CI is happy!

PR #96, opened 2024-01-18
Don't set properties on sphericart_torch target before defining the target

Luthaf:
Follow up to #95, the `sphericart_torch` needs to be defined before we can set its properties.


ceriottm:
Fixes build problems for me...

PR #105, opened 2024-01-26
Templated CUDA Kernel

nickjbrowning:
This PR includes templated CUDA code to (hopefully) allow NVCC to better optimise compute, compute_with_gradients and compute_with_hessians.


nickjbrowning:
PR failing mostly because stubs aren't functional yet.

PR #106, opened 2024-02-12
Try to build CUDA wheels

frostedoyster:


.github/workflows/build-cuda-wheels.yml
```diff
@@ -0,0 +1,98 @@
+# inspired by https://github.com/AutoGPTQ/AutoGPTQ/blob/main/.github/workflows/build_wheels_cuda.yml
+
+name: Build Python wheels with CUDA
+
+on:
+  push:
+    branches: [main]
+    tags: ["*"]
+  pull_request:
+    # Check all PR
```
Luthaf, line 10: Do we really want to build all CUDA wheels on all PR? I'm worried this will increase CI time a lot

.github/workflows/build-cuda-wheels.yml
```diff
@@ -0,0 +1,98 @@
...
+          # do not build wheels with -march=native
+          CIBW_ENVIRONMENT: SPHERICART_ARCH_NATIVE=OFF
+
+      - name: Build sphericart-torch wheels
+        run: |
+          # ensure we build the wheel from the sdist, not the checkout
+          python -m build --sdist sphericart-torch --outdir sphericart-torch/dist
+          python -m cibuildwheel sphericart-torch/dist/*.tar.gz --output-dir sphericart-torch/dist
+        env:
+          CIBW_BEFORE_ALL: bash /host/home/runner/work/sphericart/sphericart/scripts/cibw-cuda-setup.sh ${{ matrix.cuda }}
```
Luthaf, line 72: Shouldn't this also be done for the diverse sphericart (non torch) wheels?

scripts/cibw-cuda-setup.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
```
Luthaf, line 2:
```suggestion

set -eu
```

`-e` to exit if any command fails, `-u` to error on undefined bash variables

Luthaf:
Regarding the distribution of the wheels, I had an idea last week: we could have some code in Python to dynamically pick a version of the shared library (depending on CUDA presence, CUDA version and torch version for rascaline-torch). When doing a local build, there would be a single library matching the installation environment. For PyPI releases, we could use the code in this PR to build for all variations of Torch/CUDA, and then merge the resulting shared libraries in a single wheel (there are tools to extract & repack wheels).

The resulting wheel would be a bit larger, but not that bad overall since sphericart is fairly small. The other concern I have is that we would be releasing something quite different from what the developers are testing, but it might still be worth it compared to the pain of setting up an alternative PyPI index and explaining to users to use it.

.github/workflows/build-cuda-wheels.yml
frostedoyster: We don't provide any CUDA functionality for our Python (NumPy) inteface

scripts/manylinux2014_x86_64/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+ENV CUDA_PATH=/usr/local/cuda
+ENV CUDADIR=/usr/local/cuda
+
+# Verify the CUDA installation
+RUN echo "CUDA_HOME: ${CUDA_HOME}" && \
+    nvcc --version
+
+#RUN pip install torch==2.4.1+cu124 --extra-index-url https://download.pytorch.org/whl/cu124
+#RUN pip install numpy cmake
+RUN pip install torch==${PYTORCH_VERSION}+cu${CUDA_VER_NO_DOT} --extra-index-url https://download.pytorch.org/whl/cu${CUDA_VER_NO_DOT}
```
Luthaf, line 45: I don't understand why this is pulling a specific cuda build of pytorch, since we do not build the cuda code to PTX anymore. Is there a reason this is not just doing `pip install torch==${PYTORCH_VERSION}`, which will install a CUDA enabled pytorch from PyPI? 

Maybe this is not quite the case for linux-aarch64?

.github/workflows/build-torch-wheels.yml
```diff
@@ -0,0 +1,173 @@
...
+        CIBW_SKIP: "*-musllinux* *-win32 *-manylinux_i686"
+        CIBW_ARCHS: "${{ matrix.cibw-arch }}"
+        CIBW_MANYLINUX_X86_64_IMAGE: "manylinux2014_x86_64"
+        CIBW_MANYLINUX_AARCH64_IMAGE: "manylinux2014_aarch64"
+        CIBW_ENVIRONMENT: >
+          CUDA_HOME=/usr/local/cuda
+          PIP_EXTRA_INDEX_URL="https://download.pytorch.org/whl/cu${{  env.CUVERSION }}"
+          SPHERICART_ARCH_NATIVE=OFF
+        CIBW_REPAIR_WHEEL_COMMAND_LINUX: |
+          auditwheel repair --exclude libcuda.so --exclude libcuda.so.1 --exclude libc10.so --exclude libtorch.so --exclude libtorch_cpu.so --exclude libtorch_cuda.so --exclude libc10_cuda.so  --exclude libcudart.so --exclude libnvToolsExt.so  --exclude libnvrtc.so --exclude libnvrtc.so.12 -w {dest_dir} {wheel}
```
Luthaf, line 87:
```suggestion
        CIBW_REPAIR_WHEEL_COMMAND_LINUX: |
          auditwheel repair --exclude libc10.so --exclude libtorch.so --exclude libtorch_cpu.so --exclude libtorch_cuda.so --exclude libc10_cuda.so  -w {dest_dir} {wheel}
```

We should not link to anything related to cuda (since we will try to dlopen these later), so I would rather have cibuildwheel fails if we accidentally have them as `DT_NEEDED`.

.github/workflows/build-torch-wheels.yml
```diff
@@ -0,0 +1,173 @@
...
+      run: python -m pip install cibuildwheel==2.22.0
+
+    - name: Build torch wheels
+      run: python -m cibuildwheel ./sphericart-torch
+      env:
+        CUDA_HOME: /usr/local/cuda
+        PIP_EXTRA_INDEX_URL: "https://download.pytorch.org/whl/cu${{  env.CUVERSION }}"
+        CIBW_BUILD_VERBOSITY: 3
+        CIBW_BUILD: "cp${{ env.PYTHON_VER_NO_DOT }}-*"
+        CIBW_BUILD_FRONTEND: "pip; args: --no-build-isolation"
```
Luthaf, line 77:
```suggestion
        CIBW_BUILD_FRONTEND: "pip; args: --no-build-isolation --check-build-dependencies"
```

If we don't use build isolation, we should ensure that the pre-installed dependencies match what the code wants.

scripts/manylinux2014_x86_64/Dockerfile
frostedoyster: This is definitely not the case for windows, where `pip install torch` instead pulls CPU-only

Luthaf:
I guess the CI configuration now needs to be expanded to include all OS + all torch versions + more architectures?

.github/workflows/build-jax-wheels.yml
```diff
@@ -0,0 +1,86 @@
...
+    tags: ["*"]
+  pull_request:
+    # Check all PR
+
+#env:
+#  SPHERICART_NO_LOCAL_DEPS: "1"
+
+jobs:
+  build-jax-wheels:
+    runs-on: ubuntu-latest
```
Luthaf, line 15: could you give the job a name with something like `name: ${{matrix.os}} ${{matrix.arch}} Python  ${{matrix.python-version}} Torch ${{matrix.torch-version}} ...` (here and in the torch wheels build)

Luthaf: This makes it a lot easier to read when dealing with a very large list of jobs!

.github/workflows/build-jax-wheels.yml
```diff
@@ -0,0 +1,86 @@
...
+    - name: Github Actions Envs Setup
+      run: |
+        CUVERSION="${{ matrix.cuda-version }}"
+        PYTHONVERSION="${{ matrix.python-version }}"
+
+        CU_VERSION_NO_DOT=${CUVERSION//./}
+        echo CU_VERSION_NO_DOT=${CU_VERSION_NO_DOT} >> $GITHUB_ENV
+
+        CU_VERSION_DASH=${CUVERSION//./-}
+        echo CU_VERSION_DASH=${CU_VERSION_DASH} >> $GITHUB_ENV
```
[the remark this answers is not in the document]
nickjbrowning: For torch specifically, it depends if we want to support torch versions < 2.4.0, since the cu124 wheel only supports 2.4 as a minimum (https://download.pytorch.org/whl/cu124/torch). If we used cu121, we could support down to 2.1.0 (https://download.pytorch.org/whl/cu121/torch)?

.github/workflows/build-jax-wheels.yml
```diff
@@ -0,0 +1,86 @@
...
+        -t manylinux2014_"${{ matrix.arch }}" \
+        --build-arg PYTHON_VER="${{ matrix.python-version }}" \
+        --build-arg PYTHON_VER_NO_DOT="${{ env.PYTHON_VER_NO_DOT }}" \
+        --build-arg CUDA_VER="${{ matrix.cuda-version }}" \
+        --build-arg CUDA_VER_NO_DOT="${{ env.CU_VERSION_NO_DOT }}" \
+        --build-arg CUDA_VER_DASH="${{ env.CU_VERSION_DASH }}" \
+        scripts/manylinux2014_"${{ matrix.arch }}"
+    
+    - name: Install cibuildwheel
+      run: python -m pip install cibuildwheel==2.22.0
```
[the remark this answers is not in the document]
nickjbrowning: I was following pypa's documentation: https://github.com/pypa/cibuildwheel and in there their CI scripts pin versions. We can unpin if you want - up to you.

sphericart-jax/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "sphericart-jax"
 dynamic = ["version"]
-requires-python = ">=3.9"
+requires-python = ">=3.10"
```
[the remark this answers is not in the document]
nickjbrowning: i've changed both torch and jax builds to requires-python >= 3.9

sphericart/include/dynamic_cuda.hpp
```diff
@@ -366,4 +366,26 @@ static struct CUDAInitializer {
...
+inline NVRTC& GetNVRTCInstance() {
+    static NVRTC instance;
+    return instance;
+}
+
+// Macro shortcuts
+#define CUDART_INSTANCE GetCUDARTInstance()
+#define CUDA_DRIVER_INSTANCE GetCUDADriverInstance()
+#define NVRTC_INSTANCE GetNVRTCInstance() */
+
```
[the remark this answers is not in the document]
nickjbrowning: Yep!

.github/workflows/build-jax-wheels.yml
nickjbrowning: added this in!

.github/workflows/build-jax-wheels.yml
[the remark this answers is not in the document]
Luthaf: ok, that's fine then!

.github/workflows/build-wheels.yml
```diff
@@ -7,124 +7,66 @@ on:
...
     strategy:
       matrix:
-        # TODO: add windows builder
-        os: [ubuntu-20.04, macos-13]
-
+        arch: ["x86_64"]
+        python-version: ["3.12"]
+        cuda-version: ["12.1"]
+    env:
+      CIBW_SKIP: cp36-* cp37-* cp38-* cp39-* cp310-* cp311-*
```
Luthaf, line 19: I think there is a way to explicitly include just one `cpxy-`, which should be better than skipping all the ones we don't want

.github/workflows/build-jax-wheels.yml
```diff
@@ -0,0 +1,72 @@
...
+#  SPHERICART_NO_LOCAL_DEPS: "1"
+
+jobs:
+  build-jax-wheels:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: ["x86_64"]
+        python-version: ["3.10", "3.11", "3.12"]
+        cuda-version: ["12.1"]
```
Luthaf, line 20:
```suggestion
```

.github/workflows/build-torch-wheels.yml
```diff
@@ -0,0 +1,145 @@
...
+
+jobs:
+  build-torch-wheels:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: ["x86_64"]
+        python-version: ["3.12"]
+        pytorch-version: ["2.4.1", "2.5.1"]
+        cuda-version: ["12.1"]
```
Luthaf, line 21:
```suggestion
```

scripts/manylinux2014_aarch64/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+    # Remove all other Python versions
+#ENV PATH="/opt/python/cp311-cp311/bin:${PATH}"
+#RUN ln -sf /opt/python/cp311-cp311/bin/python3.11 /opt/python/cp311-cp311/bin/python
+
+# Set Python version environment variables dynamically
+ENV PATH="/opt/python/cp${PYTHON_VER_NO_DOT}-cp${PYTHON_VER_NO_DOT}/bin:${PATH}"
+RUN ln -sf /opt/python/cp${PYTHON_VER_NO_DOT}-cp${PYTHON_VER_NO_DOT}/bin/python${PYTHON_VER} /opt/python/cp${PYTHON_VER_NO_DOT}-cp${PYTHON_VER_NO_DOT}/bin/python
+
+RUN python -m ensurepip --upgrade
```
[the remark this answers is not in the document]
Luthaf: @nickjbrowning?

scripts/manylinux2014_x86_64/Dockerfile
Luthaf: @nickjbrowning?

sphericart-jax/setup.py
```diff
@@ -88,6 +102,7 @@ def run(self):
         cmdclass={
             "build_ext": cmake_ext,
             "bdist_egg": bdist_egg if "bdist_egg" in sys.argv else bdist_egg_disabled,
+            "bdist_wheel": universal_wheel,
```
Luthaf, line 105: Jax wheels are not universal, since they link to libpython through pybind11

scripts/manylinux2014_aarch64/remove_unused_python.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
nickjbrowning: All the advice I could find recommends doing it the way I have done it. I tried doing it in place but that generated a whole heap of errors.

scripts/manylinux2014_aarch64/Dockerfile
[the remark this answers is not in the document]
nickjbrowning: probably. Can remove it.

scripts/manylinux2014_aarch64/remove_unused_python.sh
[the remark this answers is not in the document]
nickjbrowning: Huh. I've just removed the scripts and it seems to work now. might have been some other issue that was creating this problem.

nickjbrowning:
@Luthaf 

https://github.com/lab-cosmo/sphericart/pull/106#discussion_r1876272542

what do you mean regarding this? We need the torch wheels specifically because we need access to the "c10/cuda" namespace, which doesn't seem to be available in cpu-torch.

Luthaf:
> which doesn't seem to be available in cpu-torch, which is what gets pulled for build if we don't specify pip_extra_index_url.

So that's the bit I don't understand. As far as I know, the linux-x64 wheels on PyPI contain all of `libtorch_cuda`, and we should be able to build against them. I would prefer it if all the things we put on PyPI are built against things that come from PyPI, since different torch build can have different ABI.

PyPI wheels might not contain anything CUDA related for linux-arm64 or windows-x64, and we will have to work around this, but for linux-x64 I'd like to build against PyPI wheel.

Luthaf:
Replaced by #160

PR #107, opened 2024-02-15
Build wheels using the abi3 tag for sphericart-jax and sphericart-torch

Luthaf:
Python offers a stable ABI since version 3.2, which can be used to build wheels targetting some minimal Python version, and use it with all future Python versions, instead of building one wheel per Python version.

This should be the right thing to do for sphericart-jax where we are linking to Python.h ourself. I'll check that the wheel work and that we don't need to pass other flags to the compiler.

I'm not sure about sphericart-torch: since we are not actually linking to Python.h, we should be able to build a wheel compatible with any Python version (but a single PyTorch version).

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -30,7 +30,7 @@ jobs:
           python-version: "3.12"
 
       - name: Install cibuildwheel
-        run: python -m pip install cibuildwheel build
+        run: python -m pip install cibuildwheel build abi3audit
 
       - name: Build sphericart wheels
         run: |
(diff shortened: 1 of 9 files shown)


Luthaf:
This should reduce the number of wheels we need to build in #106

frostedoyster:
Is this working?

Luthaf:
Not really: https://github.com/lab-cosmo/sphericart/actions/runs/7931040669/job/21654454335?pr=107

I think we need to pass a flag to pybind11 to use ABI3.

The torch wheels I'm pretty confident are not directly linking to libpython, so they should be fine with just a `py3` tag instead of a `cp38`/`cp310`/... tag.

Issue #115, opened 2024-04-29
Excessive memory usage during compilation with pip


sirmarcel:
Currently, attempting to build the `sphericart-torch` wheel with `pip` requires a large amount of RAM if many CPU cores are present. I think this is due to [this](https://github.com/lab-cosmo/sphericart/blob/fa8afd9adfe5e96ec8f4852b3723666b69d22bb0/sphericart-torch/setup.py#L52) line, which invokes `cmake` without specifying the number of jobs, which presumably will default to the total number of cores. On a HPC system those can be 40 or 80, and so compilation tends to get `kill`ed by the host OS.

While this is not catastrophic, it is inconvenient, and a waste of resources in many cases (the compilation is not much faster in parallel mode). I would suggest defaulting to some reasonable default instead, or disabling parallel builds entirely. Alternatively, the installation docs should at least mention this fact (see #116).

nickjbrowning:
Thanks for the find, this is a very good point. I'll address this in a PR tomorrow.

sirmarcel:
It's a bit suspicious. My observation is: (a) compilation dies with `kill` on the default allocation on `izar` (4GB I believe), (b) if you remove `--parallel` from the `setup.py` file of `sphericart-torch`, it works without problem, (c) requesting a node with 32GB also works, without modification.

Luthaf:
Oh, right. I can see the compiler requiring a couple of GiB per file (there are a lot of torch header to parse and template to instantiate), so parallel compilation would fail with only 4GiB of available RAM. But then the changed by @nickjbrowning would not fix it here, since the compilation would also fail with only 8 jobs.

nickjbrowning:
I've added these two environment variables to the build process:

```
SPHERICART_PARALLEL_BUILD=ON
SPHERICART_JOBS=NJOBS
```

So you can now control the number of jobs via:

```
SPHERICART_PARALLEL_BUILD=OFF pip install .[torch] #disables parallel builds
SPHERICART_JOBS=4 pip install .[torch] #uses 4 jobs for compilation
```

PR #119, opened 2024-04-29
fix to build system to limit number of threads #115

nickjbrowning:
Now uses at most 8 threads.

### DIFF
--- a/sphericart-jax/setup.py
+++ b/sphericart-jax/setup.py
@@ -10,21 +10,14 @@ from setuptools.command.build_ext import build_ext
 
 ROOT = os.path.realpath(os.path.dirname(__file__))
 SPHERICART_ARCH_NATIVE = os.environ.get("SPHERICART_ARCH_NATIVE", "ON")
-
+SPHERICART_DISABLE_PARALLEL = (
+    os.environ.get("SPHERICART_DISABLE_PARALLEL", "OFF").upper() == "ON"
+)
+SPHERICART_JOBS = os.environ.get("SPHERICART_JOBS")
 
 class cmake_ext(build_ext):
     """Build the native library using cmake"""
 
-    user_options = build_ext.user_options + [
-        ("disable-parallel-build", None, "Disable parallel build")
-    ]
-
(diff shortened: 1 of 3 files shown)


frostedoyster:
@Luthaf what do you think about this PR?
Part of me feels that Marcel should have just requested more memory on his HPC node

Luthaf:
I think this is worth it: anything that include torch takes a lot of memory to compile, and sometime the front node has limits on how much memory a single user can use. Also, we don't have that many files to compile, so using more than 8 threads won't really help much.

I'm not sure how much RAM compilation with 8 threads would take though, do we know that this is low enough to fix the issue?

nickjbrowning:
I've added two environment variables that modify the pip build process:

see: https://github.com/lab-cosmo/sphericart/issues/115#issuecomment-2225053275

frostedoyster:
Are we sure it's a good idea to have an undocumented environment variable? I think `min(cpu_count, 8)` is sufficient and simple enough

Luthaf:
>  I think min(cpu_count, 8) is sufficient 

I'm not sure. We barely have 8 files to compile, so if the build fails with `make -j` it will also fail with `make -j8`. We could do a lower default limit (2-3) if you prefer, or query available RAM and use `min(cpu_count, ram/4GiB)` (4GiB is close to what I saw a single compiler process take when building code that `#include <torch/...>`.

frostedoyster:
Querying the RAM is probably the proper fix for the original issue, so I will try to implement it

frostedoyster:
EDIT: it seems that there isn't a reliable way to get the RAM from the Python standard library, we're now doing two jobs

sphericart-jax/setup.py
```diff
@@ -1,6 +1,7 @@
 import os
 import subprocess
 import sys
+import multiprocessing
```
Luthaf, line 4:
```suggestion
```

frostedoyster: Hmmm thanks, it seems that the linter didn't catch that, I will investigate

PR #126, opened 2024-07-16
zero-neighbour patch

nickjbrowning:
This commit patches in support for zero-sized tensors for cases where there are no neighbours.

### DIFF
--- a/sphericart-torch/python/tests/test_no_points.py
+++ b/sphericart-torch/python/tests/test_no_points.py
@@ -14,34 +14,29 @@ def test_error(l_max, normalized):
 
     calculator = sphericart.torch.SphericalHarmonics(l_max=l_max, normalized=normalized)
 
-    try:
-        sph = calculator.compute(xyz)
-    except RuntimeError as e:
-        pytest.fail(f"compute threw an error: {e}")
+    sph = calculator.compute(xyz)
+    assert sph.shape == (0, (l_max + 1) ** 2)
 
-    try:
-        sph, grad_sph = calculator.compute_with_gradients(xyz)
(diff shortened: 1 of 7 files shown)


frostedoyster (commented):
Things look good for torch, I'll take care of NumPy, JAX and possibly the other APIs. Also the test failure

sphericart-torch/python/tests/test_no_points.py
```diff
@@ -0,0 +1,47 @@
...
+@pytest.mark.parametrize("l_max", [0, 3, 7, 10, 20, 50])
+def test_error(l_max, normalized):
+    xyz = torch.zeros(0, 3, dtype=torch.float64)
+
+    calculator = sphericart.torch.SphericalHarmonics(l_max=l_max, normalized=normalized)
+
+    try:
+        sph = calculator.compute(xyz)
+    except RuntimeError as e:
+        pytest.fail(f"compute threw an error: {e}")
```
Luthaf, line 20: Why are you doing this instead of just letting the exception go up the stack as usual? If you want to know which part failed, I would rather improve the original error messages

nickjbrowning: the error messages that previously resulted from passing in a zero-sized tensor are that the pointers are not initialized. Since we force the zero-sized tensor here, I thought it was appropriate to report them this way, because this is the condition that we are testing for.

nickjbrowning:
@frostedoyster your solution is much better than mine!

frostedoyster:
Ok, I will change that test and merge

frostedoyster (approved):

PR #130, opened 2024-07-28
Provide examples for spherical coordinates and/or complex spherical harmonics

frostedoyster:
As more requests are coming in for these two slightly different types of computation, it is appropriate to provide examples for these use cases

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,8 +110,8 @@ will build the documentation in a CPU-only environment.
 
 ## Other flavors of spherical harmonics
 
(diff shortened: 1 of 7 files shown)


ceriottm (approved):
LGTM. I lightly edited a couple of doc files, purely for clarity / language. 

I think this is good enough for now although I guess that users might be using sphericart mostly for derivatives and will want to see how to compute forward derivatives or how to implement a torch version to use backprop. 

We can revisit these and the decision to have as examples if and when we get requsts for this.

PR #131, opened 2024-07-30
Move compilation of CUDA code to NVRTC

nickjbrowning:


sphericart/include/jittify.hpp
Luthaf, line 1: are you still using this? Or should the file be removed?

nickjbrowning: ah yes, oops. I've removed it in latest commit.

sphericart-torch/setup.py
```diff
@@ -8,7 +8,17 @@
...
-SPHERICART_ARCH_NATIVE = os.environ.get("SPHERICART_ARCH_NATIVE", "ON")
+
+
+def get_arch_native():
+    # may not want to build processor-specific instructions
+    SPHERICART_ARCH_NATIVE = os.environ.get("SPHERICART_ARCH_NATIVE", "ON")
+    if SPHERICART_ARCH_NATIVE == "ON" and (
+        "bdist_wheel" in sys.argv or "wheel" in sys.argv
+    ):
+        return "OFF"
```
Luthaf, line 19: this looks like the wrong logic: (a) we should detect the `build_ext` step instead of the wheel one; and (b) this means we will never use `SPHERICART_ARCH_NATIVE`, even when building locally. Instead I would try to detect CIBUILDWHEELS (or however this is called), i.e. detect the environment in which we are building wheels for large scale distribution

nickjbrowning: OK well since it's not clear to me yet how we're going to be building wheels for distribution just yet (which will be a separate PR), do you think it's reasonable to instead set:

`SPHERICART_ARCH_NATIVE = os.environ.get("SPHERICART_ARCH_NATIVE", "ON")`

to OFF by default, or would this be too annoying to deal with documentation wise?

Luthaf: not sure how I missed this, sorry. Yes `SPHERICART_ARCH_NATIVE` in the environment would be fine!

nickjbrowning: I've now set SPHERICART_ARCH_NATIVE to OFF by default in all setup.py files.

.clang-format
```diff
@@ -66,7 +66,7 @@ IndentGotoLabels: true
 IndentPPDirectives: None
 IndentWidth:     4
 IndentWrappedFunctionNames: false
-InsertBraces: true
+#InsertBraces: true
```
frostedoyster, line 69: Is this important?

.gitignore
```diff
@@ -2,6 +2,9 @@ build/
 dist/
 *.egg-info
 .tox/
+tests/
 *.pyc
+*.py
+!setup.py
```
frostedoyster, line 8: These should probably be reverted (?)

sphericart-jax/setup.py
```diff
@@ -9,7 +9,7 @@
 
 
 ROOT = os.path.realpath(os.path.dirname(__file__))
-SPHERICART_ARCH_NATIVE = os.environ.get("SPHERICART_ARCH_NATIVE", "ON")
+SPHERICART_ARCH_NATIVE = os.environ.get("SPHERICART_ARCH_NATIVE", "OFF")
```
frostedoyster, line 12: We might want to keep these ON for people who compile from source so they get optimal performance

sphericart-jax/src/sphericart_jax_cuda.cu
```diff
@@ -37,7 +37,7 @@ std::unique_ptr<C<T>>& _get_or_create_sph_cuda(
 template <template <typename> class C, typename T>
 inline void cuda_sph(cudaStream_t stream, void** in, const char* opaque, std::size_t opaque_len) {
     // Parse the inputs
-    const T* xyz = reinterpret_cast<const T*>(in[0]);
+    T* xyz = reinterpret_cast<T*>(in[0]);
```
frostedoyster, line 40: Can this const be put back in now? I don't think it's a big deal either way

sphericart-jax/src/sphericart_jax_cuda.cu
```diff
@@ -58,7 +58,7 @@ inline void cuda_sph_with_gradients(
     cudaStream_t stream, void** in, const char* opaque, std::size_t opaque_len
 ) {
     // Parse the inputs
-    const T* xyz = reinterpret_cast<const T*>(in[0]);
+    T* xyz = reinterpret_cast<T*>(in[0]);
```
frostedoyster, line 61: Same here

sphericart-jax/src/sphericart_jax_cuda.cu
```diff
@@ -79,7 +79,7 @@ inline void cuda_sph_with_hessians(
...
```
frostedoyster, line 82: Same here

sphericart/include/cuda_cache.hpp
```diff
@@ -0,0 +1,450 @@
...
+
+  private:
+    CudaCacheManager() {}
+    CudaCacheManager(const CudaCacheManager&) = delete;
+    CudaCacheManager& operator=(const CudaCacheManager&) = delete;
+
+    std::unordered_map<std::string, std::unique_ptr<CachedKernel>> kernel_cache;
+};
+
+class KernelFactory {
```
Luthaf, line 399: I don't understand the difference between this and `CudaCacheManager`

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
+// DynamicCUDAHeader.hpp
+#ifndef DYNAMIC_CUDA_HEADER_HPP
+#define DYNAMIC_CUDA_HEADER_HPP
+
+#include <dlfcn.h>
+#include <stdexcept>
+#include <string>
+#include <iostream>
```
Luthaf, line 8:
```suggestion
```

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
+// DynamicCUDAHeader.hpp
+#ifndef DYNAMIC_CUDA_HEADER_HPP
+#define DYNAMIC_CUDA_HEADER_HPP
+
+#include <dlfcn.h>
```
Luthaf, line 5: this should be guarded by the right `#ifdef`

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <nvrtc.h>
+#include <any>
+
+#define NVRTC_SAFE_CALL(x)                                                                         \
+    do {                                                                                           \
+        nvrtcResult result = x;                                                                    \
+        if (result != NVRTC_SUCCESS) {                                                             \
+            std::cerr << "\nerror: " #x " failed with error "                                      \
```
Luthaf, line 20: I think this should throw an exception instead of aborting

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+        }                                                                                          \
+    } while (0)
+
+#define CUDADRIVER_SAFE_CALL(x)                                                                    \
+    do {                                                                                           \
+        CUresult result = x;                                                                       \
+        if (result != CUDA_SUCCESS) {                                                              \
+            const char* msg;                                                                       \
+            CUDADriver::instance().cuGetErrorName(result, &msg);                                   \
+            std::cerr << "\nerror: " #x " failed with error " << msg << '\n'                       \
```
Luthaf, line 34: same, this should be an exception

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+                      << "Line: " << __LINE__ << '\n';                                             \
+            exit(1);                                                                               \
+        }                                                                                          \
+    } while (0)
+
+#define CUDART_SAFE_CALL(call)                                                                     \
+    do {                                                                                           \
+        cudaError_t cudaStatus = (call);                                                           \
+        if (cudaStatus != cudaSuccess) {                                                           \
+            std::cerr << "CUDA error at " << __FILE__ << ":" << __LINE__ << " - "                  \
```
Luthaf, line 45: another exception =)

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+// Define a template to dynamically load symbols
+template <typename FuncType> FuncType load(void* handle, const char* functionName) {
+    auto func = reinterpret_cast<FuncType>(dlsym(handle, functionName));
+    if (!func) {
+        throw std::runtime_error(std::string("Failed to load function: ") + functionName);
+    }
+    return func;
+}
+
+class CUDART {
```
Luthaf, line 62: can we get some docs?

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+                      << CUDART::instance().cudaGetErrorString(cudaStatus) << '\n'                 \
+                      << "File: " << __FILE__ << '\n'                                              \
+                      << "Line: " << __LINE__ << '\n';                                             \
+            exit(1);                                                                               \
+        }                                                                                          \
+    } while (0)
+
+// Define a template to dynamically load symbols
+template <typename FuncType> FuncType load(void* handle, const char* functionName) {
+    auto func = reinterpret_cast<FuncType>(dlsym(handle, functionName));
```
Luthaf, line 55: I would add the same `#ifdef` as around `#include <dlfcn.h>`, and then `#error "platform is not supported"` on the other branch

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+    cudaMemcpy_t cudaMemcpy;
+    cudaGetErrorName_t cudaGetErrorName;
+    cudaGetErrorString_t cudaGetErrorString;
+    cudaDeviceSynchronize_t cudaDeviceSynchronize;
+    cudaPointerGetAttributes_t cudaPointerGetAttributes;
+    cudaFree_t cudaFree;
+
+  private:
+    CUDART() {
+        cudartHandle = dlopen("libcudart.so", RTLD_NOW);
```
Luthaf, line 93: Same, this should be `#ifdef linux`

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+
+        if (!cudartHandle) {
+            throw std::runtime_error(
+                "Failed to load libcudart.so. Try running \"find /usr -name libcudart.so\" and "
+                "appending the directory to your $LD_LIBRARY_PATH environment variable."
+            );
```
Luthaf, line 99:
```suggestion
            throw std::runtime_error(
                "Failed to load libcudart.so. Make sure it is available in your $LD_LIBRARY_PATH environment variable."
            );
```

The `find /usr` suggestion might be wrong on HPC, where cuda might be in modules which don't live in `/usr`

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+        cudaGetErrorString = load<cudaGetErrorString_t>(cudartHandle, "cudaGetErrorString");
+        cudaDeviceSynchronize = load<cudaDeviceSynchronize_t>(cudartHandle, "cudaDeviceSynchronize");
+        cudaPointerGetAttributes =
+            load<cudaPointerGetAttributes_t>(cudartHandle, "cudaPointerGetAttributes");
+        cudaFree = load<cudaFree_t>(cudartHandle, "cudaFree");
+    }
+
+    ~CUDART() {
+        if (cudartHandle) {
+            dlclose(cudartHandle);
```
Luthaf, line 117: `#ifdef` around this please!

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+    cuStreamDestroy_t cuStreamDestroy;
+    cuGetErrorName_t cuGetErrorName;
+    cuCtxSynchronize_t cuCtxSynchronize;
+    cuCtxPushCurrent_t cuCtxPushCurrent;
+    cuPointerGetAttribute_t cuPointerGetAttribute;
+
+  private:
+    CUDADriver() {
+        // Load CUDA driver, cuda runtime and NVRTC libraries
+        cudaHandle = dlopen("libcuda.so", RTLD_NOW);
```
Luthaf, line 187: same

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+
+        if (!cudaHandle) {
+            throw std::runtime_error(
+                "Failed to load libcuda.so. Try running \"find /usr -name libcuda.so\" and "
+                "appending the directory to your $LD_LIBRARY_PATH environment variable."
+            );
```
Luthaf, line 193:
```suggestion
            throw std::runtime_error(
                "Failed to load libcuda.so. Make sure it is available in your $LD_LIBRARY_PATH environment variable."
            );
```

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+        cuStreamDestroy = load<cuStreamDestroy_t>(cudaHandle, "cuStreamDestroy");
+        cuCtxSynchronize = load<cuCtxSynchronize_t>(cudaHandle, "cuCtxSynchronize");
+        cuGetErrorName = load<cuGetErrorName_t>(cudaHandle, "cuGetErrorName");
+        cuCtxPushCurrent = load<cuCtxPushCurrent_t>(cudaHandle, "cuCtxPushCurrent");
+        cuPointerGetAttribute = load<cuPointerGetAttribute_t>(cudaHandle, "cuPointerGetAttribute");
+    }
+
+    ~CUDADriver() {
+        if (cudaHandle) {
+            dlclose(cudaHandle);
```
Luthaf, line 226: same

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+
+  private:
+    NVRTC() {
+        nvrtcHandle = dlopen("libnvrtc.so", RTLD_NOW);
+
+        if (!nvrtcHandle) {
+            throw std::runtime_error(
+                "Failed to load libnvrtc.so. Try running \"find /usr -name libnvrtc.so\" and "
+                "appending the directory to your $LD_LIBRARY_PATH environment variable."
+            );
```
Luthaf, line 275:
```suggestion
            throw std::runtime_error(
                "Failed to load libnvrtc.so. Make sure it is available in your $LD_LIBRARY_PATH environment variable."
            );
```

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,306 @@
...
+        nvrtcGetLoweredName = load<nvrtcGetLoweredName_t>(nvrtcHandle, "nvrtcGetLoweredName");
+        nvrtcAddNameExpression =
+            load<nvrtcAddNameExpression_t>(nvrtcHandle, "nvrtcAddNameExpression");
+        nvrtcDestroyProgram = load<nvrtcDestroyProgram_t>(nvrtcHandle, "nvrtcDestroyProgram");
+        nvrtcGetErrorString = load<nvrtcGetErrorString_t>(nvrtcHandle, "nvrtcGetErrorString");
+    }
+
+    ~NVRTC() {
+        if (nvrtcHandle) {
+            dlclose(nvrtcHandle);
```
Luthaf, line 295: same

sphericart/include/macros.hpp
```diff
@@ -35,7 +35,17 @@
...
-#define cast(sph, lit) static_cast<std::remove_pointer_t<decltype(sph)>>(lit)
+template <typename T> struct remove_pointer {
+    using type = T;
+};
+
+template <typename T> struct remove_pointer<T*> {
+    using type = T;
+};
+
+template <typename T> using remove_pointer_t = typename remove_pointer<T>::type;
```
Luthaf, line 46: Could this use the version in `<type_traits>`? https://en.cppreference.com/w/cpp/types/remove_pointer

sphericart/include/templates_core.hpp
Luthaf, line 1: why does this need to be separate from the existing `template.hpp`?

sphericart/src/cuda_base.cpp
```diff
@@ -0,0 +1,257 @@
...
+        size_t end_pos = libpath.find_last_of("/\\", start_pos);
+        if (end_pos == std::string::npos) {
+            return ""; // No directory separator found
+        }
+
+        // Extract the base directory path
+        return libpath.substr(0, end_pos + 1) + directoryName;
+    }
+    return "";
+}
```
Luthaf, line 96: is this still needed?

sphericart/src/cuda_base.cpp
```diff
@@ -0,0 +1,257 @@
...
+    const int nedges,
+    const int ntotal,
+    scalar_t* xyz_grad,
+    void* cuda_stream
+) {
+
+    static const char* CUDA_CODE =
+#include "generated/wrapped_sphericart_impl.cuh"
+        ;
+    std::string kernel_name = getKernelName<scalar_t>("backward_kernel");
```
Luthaf, line 227: Should we have one file for each kernel instead of all kernels in a single file?

sphericart/src/cuda_base.cpp
```diff
@@ -0,0 +1,257 @@
...
+    int _n_total = ntotal;
+    int _nedges = nedges;
+
+    std::vector<void*> args = {&dsph, &sph_grad, &_nedges, &_n_total, &xyz_grad};
+
+    CachedKernel* kernel = kernel_factory.create(
+        kernel_name,
+        std::string(CUDA_CODE),
+        "sphericart_impl.cu",
+        {"--define-macro=CUDA_DEVICE_PREFIX=__device__"}
```
Luthaf, line 245: why can't this be hardcoded inside the `.cu` file? Do we ever define it to something different?

sphericart/src/sphericart_cuda.cpp
```diff
@@ -1,79 +1,108 @@
...
+    std::string errorMessage;
+};
+
+/*
+    This code checks whether the cuda libraries we'd need to dynamically load are available on the host
+*/
+CudaError isCudaAvailable() {
+    void* cudaHandle = dlopen("libcuda.so", RTLD_NOW);
+    void* nvrtcHandle = dlopen("libnvrtc.so", RTLD_NOW);
+    void* cudartHandle = dlopen("libcudart.so", RTLD_NOW);
```
Luthaf, line 24: this should be inside the right `#ifdef`

sphericart/src/sphericart_cuda.cpp
```diff
@@ -1,79 +1,108 @@
...
+        result.success = false;
+        result.errorMessage =
+            "Failed to load libcudart.so. Try running \"find /usr -name libcudart.so\" and "
+            "appending the directory to your $LD_LIBRARY_PATH environment variable.";
+    } else if (!nvrtcHandle) {
+        result.success = false;
+        result.errorMessage =
+            "Failed to load libnvrtc.so. Try running \"find /usr -name libnvrtc.so\" and "
+            "appending the directory to your $LD_LIBRARY_PATH environment variable.";
+    }
```
Luthaf, line 44: I don't love that we duplicate the error messages from `dynamic_cuda.hpp`. Could we use the code from there instead? Maybe inside a `try {} catch {}` block

sphericart/src/sphericart_cuda.cpp
```diff
@@ -1,79 +1,108 @@
...
 
 template <typename T> SphericalHarmonics<T>::SphericalHarmonics(size_t l_max) {
     /*
         This is the constructor of the SphericalHarmonics class. It initizlizes
        buffer space, compute prefactors, and sets the function pointers that are
        used for the actual calls
     */
+
+    checkCuda();
+
```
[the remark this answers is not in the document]
Luthaf: check that cuda is even available

Luthaf (commented):
Regarding the file organization, I'm having a bit of a hard time understanding the split between cuda.cpp, cuda_base.cpp, sphericart_cuda.hpp, templates.hpp, templates_core.hpp ...

A couple of other points:

- Could you add `SPHERICART_ARCH_NATIVE=OFF` to the existing wheel building CI?
- in `CachedKernel`, the functions names in the exception are still referirng to previous versions of the functions

sphericart/CMakeLists.txt
```diff
@@ -1,6 +1,43 @@
 cmake_minimum_required(VERSION 3.10)
 
 project(sphericart LANGUAGES C CXX)
+#R"======(, )======"
```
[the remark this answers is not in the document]
nickjbrowning: templates_core.hpp was separated from templates.hpp, since 1) we only need these specific functions in the CUDA code and 2) compilation with nvrtc will fail horribly if we try to include the original templates.hpp into the cuda_base.hpp.

cuda_base.hpp is the original front-end to the CUDA code, but now it calls nvrtc instead.

sphericart_cuda.hpp is the code which deals with the sphericart::cuda classes.

sphericart/src/cuda_base.cpp
nickjbrowning: It's easier to do all the file-preprocessing that needs to happen if we keep everything in a single file.

sphericart/src/cuda_base.cpp
nickjbrowning: it could be hardcoded there. I'll give it a try.

sphericart/src/sphericart_cuda.cpp
nickjbrowning: the reason I've done it like this is because libcuda.so is in a different directory to the CUDA SDK.

sphericart/include/sphericart_impl_i.cuh
```diff
@@ -0,0 +1,2 @@
+
+#define CUDA_DEVICE_PREFIX __device__
```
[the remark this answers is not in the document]
nickjbrowning: This was done so that this definition is set before all of the header files are included into the implementation file.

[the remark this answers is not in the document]
nickjbrowning: latest push removes the need for this.

sphericart/src/cuda_base.cpp
nickjbrowning: should be resolved in latest commit.

sphericart/src/cuda_base.cpp
Luthaf: fair enough!

.clang-format
nickjbrowning: this was just to fix an issue with my local clang version.

sphericart-torch/python/tests/test_autograd.py
```diff
@@ -18,37 +18,42 @@ def test_autograd_cartesian(xyz):
...
     if torch.cuda.is_available():
-        xyz = xyz.to(device="cuda")
-        assert torch.autograd.gradcheck(compute, xyz, fast_mode=True)
-        assert torch.autograd.gradgradcheck(compute, xyz, fast_mode=True)
+
+        xyz_cuda = xyz.detach().cuda().requires_grad_(True)
+        sph = compute(xyz_cuda)
+        sph.sum().backward()
+        assert torch.autograd.gradcheck(compute, xyz_cuda, fast_mode=True)
+        assert torch.autograd.gradgradcheck(compute, xyz_cuda, fast_mode=True)
```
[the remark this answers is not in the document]
nickjbrowning: reverted this.

sphericart-torch/python/tests/test_cuda.py
```diff
@@ -19,17 +19,19 @@ def test_cpu_vs_cuda(xyz):
...
         sph_cuda, grad_sph_cuda = calculator.compute_with_gradients(xyz_cuda)
 
-        assert torch.allclose(sph, sph_cuda.to("cpu"))
-        assert torch.allclose(grad_sph, grad_sph_cuda.to("cpu"))
+        print("--grads--")
+        print(grad_sph)
+        print(grad_sph_cuda)
+
+        assert torch.allclose(sph, sph_cuda.to("cpu"), rtol=1e-7)
+        assert torch.allclose(grad_sph, grad_sph_cuda.to("cpu"), rtol=1e-7)
```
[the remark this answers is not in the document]
nickjbrowning: reverted this.

sphericart-jax/src/sphericart_jax_cuda.cu
nickjbrowning: fixed

sphericart-jax/src/sphericart_jax_cuda.cu
nickjbrowning: fixed

sphericart-jax/src/sphericart_jax_cuda.cu
nickjbrowning: fixed

sphericart-torch/python/tests/test_cuda.py
```diff
@@ -19,17 +19,19 @@ def test_cpu_vs_cuda(xyz):
...
         sph_cuda, grad_sph_cuda = calculator.compute_with_gradients(xyz_cuda)
-
-        print(xyz[0])
         print((sph - sph_cuda.to("cpu")).flatten().max())
         print((sph - sph_cuda.to("cpu"))[0])
         print((sph - sph_cuda.to("cpu"))[1])
-        assert torch.allclose(sph, sph_cuda.to("cpu"))
-        assert torch.allclose(grad_sph, grad_sph_cuda.to("cpu"))
+        assert torch.allclose(sph, sph_cuda.to("cpu"), rtol=1e-7)
+        assert torch.allclose(grad_sph, grad_sph_cuda.to("cpu"), rtol=1e-7)
```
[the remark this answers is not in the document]
nickjbrowning: resolved.

sphericart-torch/setup.py
```diff
@@ -8,7 +8,7 @@
 
 
 ROOT = os.path.realpath(os.path.dirname(__file__))
-SPHERICART_ARCH_NATIVE = os.environ.get("SPHERICART_ARCH_NATIVE", "ON")
+SPHERICART_ARCH_NATIVE = os.environ.get("SPHERICART_ARCH_NATIVE", "OFF")
```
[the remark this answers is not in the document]
nickjbrowning: resolved

sphericart/include/cuda_base.hpp
```diff
@@ -66,29 +53,29 @@ namespace cuda {
  */
 template <typename scalar_t>
 void spherical_harmonics_cuda_base(
-    const scalar_t* __restrict__ xyz,
+    scalar_t* xyz,
```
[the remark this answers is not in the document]
nickjbrowning: resolved

sphericart/include/sphericart_cuda.hpp
```diff
@@ -71,7 +71,7 @@ template <typename T> class SphericalHarmonics {
      * @param cuda_stream Pointer to a cudaStream_t or nullptr. If this is
      * nullptr, the kernel launch will be performed on the default stream.
      */
-    void compute(const T* xyz, const size_t n_samples, T* sph, void* cuda_stream = nullptr);
+    void compute(T* xyz, const size_t n_samples, T* sph, void* cuda_stream = nullptr);
```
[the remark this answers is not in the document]
nickjbrowning: resolved

.gitignore
nickjbrowning: reverted

tox.ini
```diff
@@ -56,6 +56,7 @@ changedir = sphericart-torch
 passenv=
     PIP_EXTRA_INDEX_URL
 commands =
+    pip install {[testenv]pip_install_flags} ..
     pip install {[testenv]pip_install_flags} .
     pytest python
```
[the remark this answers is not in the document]
nickjbrowning: unsure, we should test this.

sphericart/CMakeLists.txt
```diff
@@ -15,6 +52,7 @@ OPTION(SPHERICART_OPENMP "Try to use OpenMP when compiling Sphericart" ON)
 OPTION(SPHERICART_ARCH_NATIVE "Try to use -march=native when compiling Sphericart" ON)
 OPTION(SPHERICART_ENABLE_CUDA "Are we building the CUDA backend of Sphericart?" ON)
 
+set(PACKAGE_DATA_INSTALL_DIR "package_data" CACHE PATH "Path relative to CMAKE_INSTALL_PREFIX where to additional files")
```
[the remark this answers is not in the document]
nickjbrowning: you're right - I've removed this now.

sphericart/CMakeLists.txt
```diff
@@ -67,10 +108,35 @@ endif()
 add_library(sphericart ${COMMON_SOURCES})
 
 if (CMAKE_CUDA_COMPILER AND SPHERICART_ENABLE_CUDA)
-    set_target_properties(sphericart PROPERTIES CUDA_ARCHITECTURES all)
+    #include the build/generated folder for compilation
+    set(CMAKE_INCLUDE_CURRENT_DIR ON)
```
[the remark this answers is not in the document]
nickjbrowning: I've updated this to: target_include_directories(sphericart PRIVATE ${CMAKE_CURRENT_BINARY_DIR})

sphericart/include/templates_core.hpp
nickjbrowning: Because nvcc/nvrtc is not able to compile templates.hpp, and we need to pull only a specific part of template.hpp into the cuda code, which is why I separated those specific parts out.

sphericart/src/cuda_base.cpp
nickjbrowning: nope, I've removed it.

sphericart/include/cuda_cache.hpp
```diff
@@ -0,0 +1,450 @@
...
+
+// Variadic template function to build type list
+template <typename T, typename... Ts> void buildTemplateTypes(std::string& base) {
+    base += typeName<T>(); // Add the first type
+    // If there are more types, add a comma and recursively call for the remaining types
+    if constexpr (sizeof...(Ts) > 0) {
+        base += ", ";
+        buildTemplateTypes<Ts...>(base); // Recursively call for the rest of the types
+    }
+}
```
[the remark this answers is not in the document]
nickjbrowning: I've had to revert this commit back to the original as it fails to compile with nvrtc.

sphericart/include/macros.hpp
nickjbrowning: Unfortunately not. If I use what was originally there, I get this:

```>       return self.calculator.compute(xyz)
E       RuntimeError: KernelFactory::compileAndCacheKernel: Failed to compile CUDA program:
E       sphericart_impl.cu(1480): error: namespace "std" has no member "remove_pointer_t"
E                     HARDCODED_SPH_MACRO(1, x, y, z, x2, y2, z2, buffer_sph, get_index);
E                     ^
E       
E       sphericart_impl.cu(1480): error: expected a ">"
E                     HARDCODED_SPH_MACRO(1, x, y, z, x2, y2, z2, buffer_sph, get_index);
E                     ^
E       
E       sphericart_impl.cu(1480): error: expected a "("
E                     HARDCODED_SPH_MACRO(1, x, y, z, x2, y2, z2, buffer_sph, get_index);
E                     ^
E       
E       sphericart_impl.cu(1480): error: expected an expression
E                     HARDCODED_SPH_MACRO(1, x, y, z, x2, y2, z2, buffer_sph, get_index);
E                     ^
E       
E       sphericart_impl.cu(1480): error: expected a ")"
E                     HARDCODED_SPH_MACRO(1, x, y, z, x2, y2, z2, buffer_sph, get_index);
E                     ^
E       
E       sphericart_impl.cu(1482): error: namespace "std" has no member "remove_pointer_t"
E                         HARDCODED_SPH_DERIVATIVE_MACRO(
E                         ^
E       
E       sphericart_impl.cu(1482): error: expected a ">"
E                         HARDCODED_SPH_DERIVATIVE_MACRO(
E                         ^
E       
E       sphericart_impl.cu(1482): error: expected a "("
E                         HARDCODED_SPH_DERIVATIVE_MACRO(
E                         ^
E       
E       sphericart_impl.cu(1482): error: expected an expression
E                         HARDCODED_SPH_DERIVATIVE_MACRO(
E                         ^
E       
E       sphericart_impl.cu(1482): error: expected a ")"
E                         HARDCODED_SPH_DERIVATIVE_MACRO(
E                         ^
E       
E       sphericart_impl.cu(1488): error: namespace "std" has no member "remove_pointer_t"
E                         HARDCODED_SPH_SECOND_DERIVATIVE_MACRO(
E                         ^
E       
E       sphericart_impl.cu(1488): error: expected a ">"
E                         HARDCODED_SPH_SECOND_DERIVATIVE_MACRO(
E                         ^
E       
E       sphericart_impl.cu(1488): error: expected a "("
E                         HARDCODED_SPH_SECOND_DERIVATIVE_MACRO(
E                         ^
E       
E       sphericart_impl.cu(1488): error: expected an expression
E                         HARDCODED_SPH_SECOND_DERIVATIVE_MACRO(
E                         ^
E       
E       sphericart_impl.cu(1488): error: expected a ")"
E                         HARDCODED_SPH_SECOND_DERIVATIVE_MACRO(
E                         ^
E       
E       sphericart_impl.cu(1504): error: namespace "std" has no member "remove_pointer_t"
E                     COMPUTE_SPH_L0(buffer_sph, get_index);
E                     ^
E       
E       sphericart_impl.cu(1504): error: expected a ">"
E                     COMPUTE_SPH_L0(buffer_sph, get_index);
E                     ^
E       
E       sphericart_impl.cu(1504): error: expected a "("
E                     COMPUTE_SPH_L0(buffer_sph, get_index);
E                     ^
E       
E       sphericart_impl.cu(1504): error: expected an expression
E                     COMPUTE_SPH_L0(buffer_sph, get_index);
E                     ^
E       
E       sphericart_impl.cu(1504): error: expected a ")"
E                     COMPUTE_SPH_L0(buffer_sph, get_index);
E                     ^
E       
E       sphericart_impl.cu(1506): error: namespace "std" has no member "remove_pointer_t"
E                         COMPUTE_SPH_DERIVATIVE_L0(
E                         ^
E       
E       sphericart_impl.cu(1506): error: expected a ">"
E                         COMPUTE_SPH_DERIVATIVE_L0(
E                         ^
E       
E       sphericart_impl.cu(1506): error: expected a "("
E                         COMPUTE_SPH_DERIVATIVE_L0(
E                         ^
E       
E       sphericart_impl.cu(1506): error: expected an expression
E                         COMPUTE_SPH_DERIVATIVE_L0(
E                         ^
E       
E       sphericart_impl.cu(1506): error: expected a ")"
E                         COMPUTE_SPH_DERIVATIVE_L0(
E                         ^
E       
E       sphericart_impl.cu(1511): error: namespace "std" has no member "remove_pointer_t"
E                             COMPUTE_SPH_SECOND_DERIVATIVE_L0(
E                             ^
E       
E       sphericart_impl.cu(1511): error: expected a ">"
E                             COMPUTE_SPH_SECOND_DERIVATIVE_L0(
E                             ^
E       
E       sphericart_impl.cu(1511): error: expected a "("
E                             COMPUTE_SPH_SECOND_DERIVATIVE_L0(
E                             ^
E       
E       sphericart_impl.cu(1511): error: expected an expression
E                             COMPUTE_SPH_SECOND_DERIVATIVE_L0(
E                             ^
E       
E       sphericart_impl.cu(1511): error: expected a ")"
E                             COMPUTE_SPH_SECOND_DERIVATIVE_L0(
E                             ^
E       
E       30 errors detected in the compilation of "sphericart_impl.cu".

../../miniconda3/envs/CSCS/lib/python3.10/site-packages/sphericart/torch/__init__.py:176: RuntimeError```

File: sphericart/include/cuda_cache.hpp (hunk 8 above)
Comment (nickjbrowning) [reply sphericart/include/cuda_cache.hpp:380]: I've merged this class into KernelFactory.

File: sphericart/src/sphericart_cuda.cpp (hunk 29 above)
Comment (nickjbrowning) [reply sphericart/src/sphericart_cuda.cpp:44]: i've modified this now so that the class constructors return whether or not their respective handles are non-zero.

File: sphericart/include/cuda_cache.hpp (hunk 42 above)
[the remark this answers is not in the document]
Comment (nickjbrowning) [reply sphericart/include/cuda_cache.hpp:45]: RuntimeError: KernelFactory::compileAndCacheKernel: Failed to compile CUDA program:
E       __nv_name_map(2): error: identifier "f" is undefined
E         #pragma nv_mangled_name spherical_harmonics_kernel<f>

this is the error, specifically.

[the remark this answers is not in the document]
Comment (nickjbrowning) [reply sphericart/include/cuda_cache.hpp:45]: There might be a way around this using nvrtcGetTypeName. I'll try it out tomorrow.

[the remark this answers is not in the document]
Comment (nickjbrowning) [reply sphericart/include/cuda_cache.hpp:45]: Nevermind... according to the docs: https://docs.nvidia.com/cuda/nvrtc/index.html

nvrtcGetTypeName just does it in virtually the same way I am doing it here (see https://github.com/tpn/cuda-samples/blob/master/v8.0/include/nvrtc.h#L529). Additionally, nvrtcGetTypeName is a templated function so cannot be loaded at runtime.

**Comment by ceriottm** [comment]
I can build on my laptop (no GPU) and on my workstation (old P1000, cuda 12.4)

File: sphericart/include/cuda_base.hpp [hunk 43]
```diff
@@ -66,29 +53,29 @@ namespace cuda {
...
 template <typename scalar_t>
 void spherical_harmonics_backward_cuda_base(
-    const scalar_t* __restrict__ dsph,
-    const scalar_t* __restrict__ sph_grad,
+    scalar_t* dsph,
+    scalar_t* sph_grad,
     const int nedges,
     const int ntotal,
-    scalar_t* __restrict__ xyz_grad,
+    scalar_t* xyz_grad,
```
frostedoyster, line 78: I see that the consts were added back to the forward equivalent of this. Would it be possible to do the same here? (Not high priority, I see that the API is now back to const)

tox.ini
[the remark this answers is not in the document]
frostedoyster: You should be able to test it with
`python -m pip install tox`
`python -m tox` (in the sphericart repo)
on a GPU machine

frostedoyster:
I also see some compilation warnings:
```
      In file included from /home/filippo/code/sphericart/sphericart-jax/sphericart/src/cuda_base.cpp:4:
      /home/filippo/code/sphericart/sphericart-jax/sphericart/include/cuda_cache.hpp: In member function ‘void CachedKernel::checkAndAdjustSharedMem(int)’:
      /home/filippo/code/sphericart/sphericart-jax/sphericart/include/cuda_cache.hpp:183:22: warning: unused variable ‘res’ [-Wunused-variable]
        183 |             CUresult res = driver.cuCtxGetDevice(&cuDevice);
            |                      ^~~
      /home/filippo/code/sphericart/sphericart-jax/sphericart/src/cuda_base.cpp: In function ‘size_t total_buffer_size(size_t, size_t, size_t, size_t, bool, bool)’:
      /home/filippo/code/sphericart/sphericart-jax/sphericart/src/cuda_base.cpp:22:12: warning: unused parameter ‘GRID_DIM_X’ [-Wunused-parameter]
         22 |     size_t GRID_DIM_X,
            |     ~~~~~~~^~~~~~~~~~
```

nickjbrowning:
> I also see some compilation warnings:
> 
> ```
>       In file included from /home/filippo/code/sphericart/sphericart-jax/sphericart/src/cuda_base.cpp:4:
>       /home/filippo/code/sphericart/sphericart-jax/sphericart/include/cuda_cache.hpp: In member function ‘void CachedKernel::checkAndAdjustSharedMem(int)’:
>       /home/filippo/code/sphericart/sphericart-jax/sphericart/include/cuda_cache.hpp:183:22: warning: unused variable ‘res’ [-Wunused-variable]
>         183 |             CUresult res = driver.cuCtxGetDevice(&cuDevice);
>             |                      ^~~
>       /home/filippo/code/sphericart/sphericart-jax/sphericart/src/cuda_base.cpp: In function ‘size_t total_buffer_size(size_t, size_t, size_t, size_t, bool, bool)’:
>       /home/filippo/code/sphericart/sphericart-jax/sphericart/src/cuda_base.cpp:22:12: warning: unused parameter ‘GRID_DIM_X’ [-Wunused-parameter]
>          22 |     size_t GRID_DIM_X,
>             |     ~~~~~~~^~~~~~~~~~
> ```

this should be fixed now.

sphericart/include/cuda_base.hpp
nickjbrowning: this should be resolved now.

sphericart-jax/src/sphericart_jax_cuda.cpp
```diff
@@ -2,38 +2,116 @@
...
+template <template <typename> class C, typename T>
+using CacheMapCUDA = std::map<size_t, std::unique_ptr<C<T>>>;
+
+template <template <typename> class C, typename T>
+std::unique_ptr<C<T>>& _get_or_create_sph_cuda(size_t l_max) {
+    // Static map to cache instances based on parameters
+    static CacheMapCUDA<C, T> sph_cache;
+    static std::mutex cache_mutex;
+
+    // Check if instance exists in cache, if not create and store it
```
Luthaf, line 31: @frostedoyster should we purge the instances after some point? Or do we let this map grow unbounded? (This would be for a separate PR!)

sphericart-torch/python/tests/test_autograd.py
```diff
@@ -18,37 +18,37 @@ def test_autograd_cartesian(xyz):
...
         return sph
 
     assert torch.autograd.gradcheck(compute, xyz, fast_mode=True)
     assert torch.autograd.gradgradcheck(compute, xyz, fast_mode=True)
 
     if torch.cuda.is_available():
-        xyz = xyz.to(device="cuda")
-        assert torch.autograd.gradcheck(compute, xyz, fast_mode=True)
-        assert torch.autograd.gradgradcheck(compute, xyz, fast_mode=True)
+        xyz_cuda = xyz.detach().cuda().requires_grad_(True)
```
Luthaf, line 30:
```suggestion
        xyz_cuda = xyz_in.detach().cuda().requires_grad_(True)
```

right?

sphericart-torch/python/tests/test_autograd.py
```diff
@@ -18,37 +18,37 @@ def test_autograd_cartesian(xyz):
...
```
Luthaf, line 49:
```suggestion
        xyz_cuda = xyz_in.detach().cuda().requires_grad_(True)
```

sphericart/CMakeLists.txt
```diff
@@ -67,10 +103,34 @@ endif()
...
+    #include the build/generated folder for compilation
+    target_include_directories(sphericart PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+    target_include_directories(sphericart PRIVATE ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
+    target_link_libraries(sphericart PRIVATE ${CMAKE_DL_LIBS})
+
+    set(SPHERICART_CUDA_HEADERS
+        "include/sphericart_impl.cuh"
+        "include/templates_core.hpp"
+        "include/macros.hpp"
+
```
Luthaf, line 115:
```suggestion
```

sphericart/CMakeLists.txt
```diff
@@ -2,6 +2,39 @@ cmake_minimum_required(VERSION 3.10)
 
 project(sphericart LANGUAGES C CXX)
 
+function(make_includeable INPUT_FILE OUTPUT_FILE)
```
Luthaf, line 5: Could we get a comment explaining this function?

sphericart/CMakeLists.txt
```diff
@@ -2,6 +2,39 @@ cmake_minimum_required(VERSION 3.10)
...
+        message(FATAL_ERROR "Error: The input file '${INPUT_FILE}' does not exist.")
+    endif()
+    file(READ ${INPUT_FILE} content)
+    # Format the content to be included as a raw string in C++
+    set(content "R\"======(\n${content}\n)======\"")
+    # Write the formatted content to the output file
+    file(WRITE ${OUTPUT_FILE} "${content}")
+endfunction()
+
+function(prepend_headers_to_source INPUT_FILE OUTPUT_FILE HEADERS)
```
Luthaf, line 16: Same here

sphericart/include/cuda_cache.hpp
```diff
@@ -0,0 +1,432 @@
...
+
+    std::string kernel_name;
+    std::string kernel_code;
+    std::string source_name;
+    std::vector<std::string> options;
+};
+
+/*
+Factory class to create and store compiled cuda kernels for caching as a simple name-based hashmap.
+ALlows both compi.ing from a source file, or for compiling from a variable containing CUDA code.
```
Luthaf, line 360:
```suggestion
Allows both compiling from a source file, or for compiling from a variable containing CUDA code.
```

sphericart/include/cuda_cache.hpp
```diff
@@ -0,0 +1,432 @@
...
+        this->options = options;
+    }
+
+    CachedKernel() = default;
+
+    void set(CUmodule m, CUfunction f, CUcontext c) {
+        this->module = m;
+        this->function = f;
+        this->context = c;
+    }
```
Luthaf, line 99: This feels weird as a public function. Should it be removed and inlined where used?

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,357 @@
...
+#endif
+#include <stdexcept>
+#include <string>
+#include <functional>
+#include <unordered_map>
+#include <cuda.h>
+#include <cuda_runtime.h>
+#include <nvrtc.h>
+#include <any>
+#include <sstream>
```
Luthaf, line 18:
```suggestion
#include <stdexcept>
#include <string>
#include <functional>
#include <unordered_map>
#include <any>
#include <sstream>

#include <cuda.h>
#include <cuda_runtime.h>
#include <nvrtc.h>
```
Just for clarity

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    // Prevent copying
+    CUDART(const CUDART&) = delete;
+    CUDART& operator=(const CUDART&) = delete;
+
+    void* cudartHandle = nullptr;
+};
+
+class CUDADriver {
```
Luthaf, line 137: Documentation please

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    // Prevent copying
+    CUDADriver(const CUDADriver&) = delete;
+    CUDADriver& operator=(const CUDADriver&) = delete;
+
+    void* cudaHandle = nullptr;
+};
+
+class NVRTC {
```
Luthaf, line 247: Documentation please

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,357 @@
...
+don't want these classes destroyed before jax or pytorch classes/objects are.
+*/
+
+// Static memory buffer for each class
+static std::aligned_storage<sizeof(CUDART), alignof(CUDART)>::type cudartBuffer;
+static std::aligned_storage<sizeof(CUDADriver), alignof(CUDADriver)>::type cudaDriverBuffer;
+static std::aligned_storage<sizeof(NVRTC), alignof(NVRTC)>::type nvrtcBuffer;
+
+// global references, use inline instead of extern as we want to define the implementation here
+inline CUDART& cudart = reinterpret_cast<CUDART&>(cudartBuffer);
```
Luthaf, line 330: does inline do anything for global variables? I thought it only applied to functions.

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,357 @@
...
+static int nifty_counter = 0;
+
+// Counter class for initializing and destroying static objects
+static struct CUDAInitializer {
+
+    CUDAInitializer() {
+        if (nifty_counter++ == 0) {
+            new (&cudart) CUDART();
+            new (&cudadriver) CUDADriver();
+            new (&nvrtc) NVRTC();
```
Luthaf, line 343: I'm not sure I understand how this will work on systems without a CUDA SDK: the os will try to create these when loading `sphericart.so`, fail to load `libcuda.so` and crash with the errors above

sphericart/src/sphericart_cuda.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include "sphericart_cuda.hpp"
+
+using namespace sphericart::cuda;
+
+/*
+    This code checks whether the cuda libraries we'd need to dynamically load are available on the host
+*/
+
+void checkCuda() {
+    if (!cudadriver.loaded()) {
```
Luthaf, line 18: with the current version of the code, I don't see how this will ever be `false`

sphericart/include/dynamic_cuda.hpp
```diff
@@ -0,0 +1,357 @@
...
+
+// Static memory buffer for each class
+static std::aligned_storage<sizeof(CUDART), alignof(CUDART)>::type cudartBuffer;
+static std::aligned_storage<sizeof(CUDADriver), alignof(CUDADriver)>::type cudaDriverBuffer;
+static std::aligned_storage<sizeof(NVRTC), alignof(NVRTC)>::type nvrtcBuffer;
+
+// global references, use inline instead of extern as we want to define the implementation here
+inline CUDART& cudart = reinterpret_cast<CUDART&>(cudartBuffer);
+inline CUDADriver& cudadriver = reinterpret_cast<CUDADriver&>(cudaDriverBuffer);
+inline NVRTC& nvrtc = reinterpret_cast<NVRTC&>(nvrtcBuffer);
```
Luthaf, line 332:
```suggestion
inline CUDART& CUDART_INSTANCE = reinterpret_cast<CUDART&>(cudartBuffer);
inline CUDADriver& CUDA_DRIVER_INSTANCE = reinterpret_cast<CUDADriver&>(cudaDriverBuffer);
inline NVRTC& NVRTC_INSTANCE = reinterpret_cast<NVRTC&>(nvrtcBuffer);
```

just to make it clear these are global-ish variables

sphericart/src/sphericart_cuda.cpp
```diff
@@ -0,0 +1,188 @@
...
+        throw std::runtime_error(
+            "Failed to load libcuda.so. Try running \"find /usr -name libcuda.so\" and "
+            "appending the directory to your $LD_LIBRARY_PATH environment variable."
+        );
```
Luthaf, line 22: This is duplicated from the `CUDA` class constructor, I would prefer if the same error message did not exist twice in the same code base.

tox.ini
```diff
@@ -99,12 +99,15 @@ deps =
...
-    jax[cpu]
     pytest
 
 passenv=
     PIP_EXTRA_INDEX_URL
 commands =
+    # Install this one manually. Listing it in the deps list above does not install jaxlib.
+    # Note: jax[cuda12] is not available on Windows and MacOS.
+    bash -c 'command -v nvcc && python -m pip install -U "jax[cuda12]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html || python -m pip install -U jax[cpu]'
+
```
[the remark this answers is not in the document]
frostedoyster: I don't see the immediate downside and it does allow for catching more bugs on the CSCS CI

sphericart-torch/python/tests/test_autograd.py
nickjbrowning: no, xyz_in is a parameter of the nested function compute. autograd.gradcheck takes that function in as a parameter and we pass in xyz_cuda as a variable.

sphericart-torch/python/tests/test_autograd.py
nickjbrowning: same as above

sphericart/include/cuda_cache.hpp
nickjbrowning: I've removed it.

sphericart/include/dynamic_cuda.hpp
nickjbrowning: since c++17 (i think) you can inline a variable, which allows you to define it identically with external linkage across multiple translation units. Since we're running from a single header source (which I'd like to keep) this is the way to handle this I think.

sphericart/include/dynamic_cuda.hpp
nickjbrowning: you are right!

maybe we should "do nothing" if the libraries aren't found - no warnings etc, and then rely on the sphericart_cuda.cpp classes to throw the runtime errors, that will allow us to do both CPU and GPU then.

sphericart/src/sphericart_cuda.cpp
nickjbrowning: this should be fixed now with the changes above. It doesn't return false, but will throw errors if the libs aren't found.

sphericart/include/dynamic_cuda.hpp
nickjbrowning: fixed!

sphericart/src/sphericart_cuda.cpp
nickjbrowning: this is resolved now, only sphericart_cuda.cpp will throw this error.

sphericart/include/dynamic_cuda.hpp
nickjbrowning: I've also added a brief comment on the remove_pointer structs

tox.ini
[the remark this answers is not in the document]
ceriottm: Without this, it won't run the cuda code through jax, and we woul have indeed missed the bugs from last week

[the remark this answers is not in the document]
Luthaf: But there is a separate test section above which is already using cuda, this is for the examples. If there is stuff in there that acts like a test, we should maybe move it to the tests instead.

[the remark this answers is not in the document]
Luthaf: But this is fine for now, we can revisit later if this becomes a problem.

sphericart/include/dynamic_cuda.hpp
Luthaf: Well, today I learned! This is useful, but the syntax is a bit confusing …

sphericart-torch/python/tests/test_autograd.py
Luthaf: you're right, I got confused here

sphericart/src/sphericart_cuda.cpp
```diff
@@ -0,0 +1,188 @@
...
+/*
+    This code checks whether the cuda libraries we'd need to dynamically load are available on the host
+*/
+
+void checkCuda() {
+    if (!CUDA_DRIVER_INSTANCE.loaded()) {
+        throw std::runtime_error(
+            "Failed to load libcuda.so. Try running \"find /usr -name libcuda.so\" and "
+            "appending the directory to your $LD_LIBRARY_PATH environment variable."
+        );
```
Luthaf, line 22: OK, nice that this is now the only place with an error mesage! Could it be updated to not reference `/usr` since `libcuda` can be elsewhere? This was already done on the other messages but these fell through!

nickjbrowning: I've updated it, do you agree with the new message?

Luthaf: yes, looks a lot better!

Luthaf (approved):
Thanks a lot for your work @nickjbrowning!

Issue #98, opened 2024-01-19
Make APIs as uniform as possible


frostedoyster:
At the moment, the APIs for `C/C++/NumPy/torch`, `JAX`, `Julia` and `CUDA` are all slightly different. We should discuss up to what point we should aim at making them uniform and where we should instead give way to the idioms of each language/framework

cortner:
Im open for anything really since at our end we have to write wrappers anyhow to make this compatible with how we organize computations.

Luthaf:
We had a discussion about this today, here is a quick summary:

- use separate classes/objects for spherical harmonics and solid harmonics instead of a "normalize" parameter
- use functional-looking API for most things, using a custom `__call__` for numpy and torch

### Julia

Currently does something like this
```julia
basis = SolidHarmonics(10)

# one of
sph = basis(R)
sph = compute(basis, R)
```

### Jax

Currently does something like this
```py
sph = compute_spherical_harmonics(lmax=10, R, normalize=True)
```

We can change it to
```py
calculator = SphericalHarmonics(lmax=10)
sph = compute_spherical_harmonics(calculator, R)
sph = calculator(R)

calculator = SolidHarmonics(lmax=10)
sph = compute_solid_harmonics(calculator, R)
sph = calculator(R)
```

### Torch/Numpy

We can change these to
```py
calculator = SphericalHarmonics(lmax=10)
sph = calculator(R)

# add this one
calculator = SolidHarmonics(lmax=10)
sph = calculator(R)
```

sirmarcel:
Also would be nice to be very explicit about which part of the "what sphericart computes" article maps to which arguments/classes. Currently it's technically written down, but a bit hard to parse.

cortner:
My only gripe with what you suggest is that the type of the calculator should determine whether it's spherical or solids - hence just "compute".

But as a general rule I'm not certain that unifying too much is even a good thing. Different languages and different frameworks make different usage conventions natural.

sirmarcel:
For `jax`, I'd suggest following the [`e3x` API](https://e3x.readthedocs.io/stable/_autosummary/e3x.so3.irreps.spherical_harmonics.html#e3x.so3.irreps.spherical_harmonics), i.e., just having functions `solid_harmonics` and `spherical_harmonics`. `compute_spherical_harmonics` accepting some opaque calculator object as argument doesn't seem very intuitive and probably doesn't play nicely with `jit`.

[...]

PR #137, opened 2024-08-18
Change `normalized` to `SphericalHarmonics` and `SolidHarmonics` across all APIs

frostedoyster:
This is much more intuitive for the user (at the moment, if they ask for the spherical harmonics, they get something else by default), matches the Julia API, and also seems to be accepted in the field (see e3x API for spherical and solid harmonics)

### DIFF
--- a/examples/pytorch/example.py
+++ b/examples/pytorch/example.py
@@ -19,7 +19,10 @@ class SHModule(torch.nn.Module):
     `torch.nn.Module`"""
 
     def __init__(self, l_max, normalized=False):
-        self.spherical_harmonics = sphericart.torch.SphericalHarmonics(l_max, normalized)
+        if normalized:
+            self.spherical_harmonics = sphericart.torch.SphericalHarmonics(l_max)
+        else:
+            self.spherical_harmonics = sphericart.torch.SolidHarmonics(l_max)
         super().__init__()
 
     def forward(self, xyz):
@@ -43,7 +46,10 @@ def sphericart_example(l_max=10, n_samples=10000, normalized=False):
 
     # ===== API calls =====
 
-    sh_calculator = sphericart.torch.SphericalHarmonics(l_max, normalized=normalized)
+    if normalized:
+        sh_calculator = sphericart.torch.SphericalHarmonics(l_max)
+    else:
+        sh_calculator = sphericart.torch.SolidHarmonics(l_max)
 
     # the interface allows to return directly the forward derivatives (up to second
     # order), similar to the Python version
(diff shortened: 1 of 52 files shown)


frostedoyster:
I can't get torch to work, I think the inheritance `torch::CustomClassHolder` -> `SphericalHarmonics` -> `SolidHarmonics` is breaking the `CustomClassHolder` mechanics.
Multiple inheritance of `SolidHarmonics` from `SphericalHarmonics` and  `CustomClassHolder` doesn't compile (understandably perhaps).
@Luthaf please have a look if and when you can

frostedoyster:
This would solve #98

Luthaf:
> I can't get torch to work, I think the inheritance torch::CustomClassHolder -> SphericalHarmonics -> SolidHarmonics is breaking the CustomClassHolder mechanics.
Multiple inheritance of SolidHarmonics from SphericalHarmonics and CustomClassHolder doesn't compile (understandably perhaps).

Yes, you'll need multiple classes at the Torch API level, one for each calculator.

frostedoyster:
Torch should be good. The docs are not ideal for now, but I'm planning to do a thorough docs revision PR after #139 is fixed

examples/pytorch/example.py
```diff
@@ -19,7 +19,10 @@ class SHModule(torch.nn.Module):
     `torch.nn.Module`"""
 
     def __init__(self, l_max, normalized=False):
-        self.spherical_harmonics = sphericart.torch.SphericalHarmonics(l_max, normalized)
+        if normalized:
+            self.spherical_harmonics = sphericart.torch.SphericalHarmonics(l_max)
+        else:
+            self.spherical_harmonics = sphericart.torch.SolidHarmonics(l_max)
```
Luthaf, line 25: Does it make sense to keep both options here now there there is no longer a normalized option? I would only demonstrate one of the calculator in the example and not muddy the water with more option than required

sphericart/include/sphericart.h
```diff
@@ -13,8 +13,10 @@
 
 #include "sphericart.hpp"
 
-using sphericart_calculator_t = sphericart::SphericalHarmonics<double>;
-using sphericart_calculator_f_t = sphericart::SphericalHarmonics<float>;
+using sphericart_spherical_harmonics_calculator_t = sphericart::SphericalHarmonics<double>;
+using sphericart_spherical_harmonics_calculator_f_t = sphericart::SphericalHarmonics<float>;
+using sphericart_solid_harmonics_calculator_t = sphericart::SolidHarmonics<double>;
+using sphericart_solid_harmonics_calculator_f_t = sphericart::SolidHarmonics<float>;
```
Luthaf, line 19: Do we really need separate types here? We could reduce the number of functions by either having a single calculator type, which is initialized to be either `SphericalHarmonics` or `SolidHarmonics`, or having two calculator types by a single function that can take both and call the right function on them.

If this is not an issue, we could also keep the two separate, but I feel like we have a lot of functions in the C API for no reason.

examples/pytorch/example.py
frostedoyster: Yes, let's just keep the SphericalHarmonics

sphericart/include/sphericart.h
frostedoyster: I agree that we have a lot of functions (and we would have even more if we wanted to have a CUDA C API for e.g. CuPy). However, I feel like this is in line with the other APIs and the original idea behind this PR was to make the APIs as uniform as possible

Luthaf (approved):

sphericart/include/sphericart.h
frostedoyster: Thanks. And yes, let's discuss in person

Issue #139, opened 2024-08-20
Remove weird `sphericart.torch` mechanics for double backward


frostedoyster:
In torch, there is no way to know if a second derivative call might be executed by the user (unlike for the first derivative, where `requires_grad` can be checked).
As a result, in the current API, we require the user to specify if second derivatives will be used at class initialization. This is pretty useless for two reasons:
- if the model is torchscripted, the class can't be re-initialized so one is stuck with first derivatives only
- the model fails **silently** if second derivatives are called and they were not requested at initialization

The only way I see to make this feature usable is to calculate the second derivatives on the fly when their calculation is needed. This will recompute the values and first derivatives of the spherical harmonics, but the current approach which avoids the recomputation is unsustainable in practice. We should also find a way to mark the second derivative function as non-differentiable to avoid, once again, silent failures if people try to differentiate 3 or more times. Something similar to `@once_differentiable` (https://discuss.pytorch.org/t/what-does-the-function-wrapper-once-differentiable-do/31513), but for C++ torch.

ceriottm:
Maybe having a separate class for second derivatives?

frostedoyster:
This is partially fixed thanks to warnings.
Once we can reasonably expect torch 2.4 as a minimum requirement, we will be able to fix it once and for all with
https://pytorch.org/tutorials/advanced/cpp_custom_ops.html

Luthaf:
I'm not sure I see how the link you shared would fix the issue? By allowing to define backward Python-side?

frostedoyster:
This will be fixed by #219

frostedoyster:
Fixed in #219

PR #141, opened 2024-08-27
Raise warning for torch double backward

frostedoyster:
After discussing #139, @Luthaf and I came to the conclusion that we can't do much without sacrificing a lot of performance in the important case where only mixed derivatives are needed. Therefore, we're adding a warning and thoroughly documenting the potential issues that can arise


Luthaf (approved):

Luthaf:
The changes looks good but CI failure is strange here

ceriottm:
> The changes looks good but CI failure is strange here

Guess you'll have to test this - neither Filippo nor I have eaten from the tree of vendor lock-in

ceriottm:
ok I reverted all my foolish attempts to fix the CI

Luthaf:
I can not reproduce the issue locally, but I just realized that CI was still using macos-13 runner, which is x86_64. Torch is no longer supporting this architecture, so I switched to macos-14 runner, which is arm64. If this fixes CI I think we are good to go, this could be an architecture-specific bug, or a bug with older versions of torch.

frostedoyster:
Thanks for the fix!

PR #143, opened 2024-08-28
Documentation review

frostedoyster:
Attempt to update and clean up the documentation.
Preview: https://sphericart--143.org.readthedocs.build/en/143/


frostedoyster:
I haven't touched the CUDA C++ and torch docs yet. I think we should wait for CUDA C++ to match the C++ API and for #141 to be merged. Everything else should be ready

ceriottm:
The CUDA c++ example seems to imply that there is just `compute`, but the API also specify `compute_with_gradient` and `compute_with_hessian` - I suspect it needs updating.

ceriottm (approved):

PR #150, opened 2024-09-06
Update benchmarks to new API

ceriottm:
Benchmarks were left broken by the change in API for jax and pytorch. Now everything should be compatible.


frostedoyster (approved):
Thank you!

PR #152, opened 2024-10-11
Cache only once in sphericart-jax

frostedoyster:
In `main`, this code:

```
import jax
import sphericart.jax

# set jax to cpu
# jax.config.update("jax_platform_name", "cpu")


key = jax.random.PRNGKey(0)
xyz = 6 * jax.random.normal(key, (100, 3))

function = sphericart.jax.spherical_harmonics

sph = function(xyz, l_max=3)
hessian = jax.hessian(lambda xyz: jax.numpy.sum(function(xyz, l_max=3)))(xyz)
grad = jax.grad(lambda xyz: jax.numpy.sum(function(xyz, l_max=3)))(xyz)
hessian = jax.hessian(lambda xyz: jax.numpy.sum(function(xyz, l_max=3)))(xyz)
sph = function(xyz, l_max=3)
sph = function(xyz, l_max=3)
```

creates 3 classes: one for sph, one for dsph, one for ddsph.
This PR fixes it

### DIFF
--- a/sphericart-jax/src/sphericart_jax_cpu.cpp
+++ b/sphericart-jax/src/sphericart_jax_cpu.cpp
@@ -21,9 +21,11 @@ template <template <typename> class C, typename T>
 using CacheMapCPU = std::map<size_t, std::unique_ptr<C<T>>>;
 
 template <template <typename> class C, typename T>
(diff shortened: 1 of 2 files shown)


Luthaf (approved):
This makes sense: we only want one cache on CPU, not that every function has its own cache.

PR #156, opened 2024-12-20
Try a more typical cibuildwheel setup

Luthaf:


Luthaf:
This should be a more typical cibuildwheels setup, with fewer customization.

frostedoyster (approved):

PR #158, opened 2025-01-06
`metatensor` interface

frostedoyster:
A `metatensor` interface for `sphericart`


Luthaf (commented):
The code looks good, but this might be worth a bit more documentation, both adding a link to metatensor in the class docstring and putting some reference to these new classes in the main documentation.

frostedoyster:
Unfortunately it wasn't possible to share code between the two interfaces because `sphericart-torch` doesn't depend on `sphericart` (and I think it should stay this way)

examples/metatensor/example.py
```diff
@@ -0,0 +1,54 @@
...
+
+l_max = 15
+n_samples = 100
+
+xyz = TensorMap(
+        keys=Labels.single(),
+        blocks=[
+            TensorBlock(
+                values=np.random.rand(n_samples, 3, 1),
+                samples=Labels(
```
Luthaf, line 16: should we indicate somewhere (maybe here and in the docstring?) that the output will have the same samples as the input?

PR #160, opened 2025-01-10
Updates to the wheels building process

Luthaf:
This is a rebased version of #106, to be rebase-merged down the line & keep all the contributions intact.

In this PR, we added support for cibuildwheels to sphericart-torch and sphericart-jax, removing the need for end users to compile the code themself. 

We build wheels for these on arm64-macos and x86_64-linux (not x86_64-macos, since it is no longer supported by torch and jax). The linux wheels are CUDA-enabled, and can run with any CUDA version thanks to #131. We build a single sphericart-torch wheel for all Python version and PyTorch versions; and one sphericart-jax wheel per Python version.


sphericart-torch/python/sphericart/torch/__init__.py
```diff
@@ -2,62 +2,74 @@
...
+        elif sys.platform.startswith("linux"):
+            path = os.path.join(expected_prefix, "lib", "libsphericart_torch.so")
+        elif sys.platform.startswith("win"):
+            path = os.path.join(expected_prefix, "bin", "sphericart_torch.dll")
+        else:
+            raise ImportError("Unknown platform. Please edit this file")
+
+        if os.path.isfile(path):
+            # if windows:
+            #     _check_dll(path)
```
frostedoyster, line 46: Probably dead code

frostedoyster (approved):

PR #165, opened 2025-01-16
Release v1.0.0-rc2

Luthaf:


Luthaf:
CSCS CI is reporting the job as stuck: 

> This job is stuck because of one of the following problems. There are no active runners online, no runners for the protected branch , or no runners that match all of the job's tags

So I'll merge without it

PR #166, opened 2025-01-16
CUDA Toolkit version + Jax incompatibility check

nickjbrowning:


sphericart-jax/python/sphericart/jax/__init__.py
```diff
@@ -1,18 +1,73 @@
...
+    jax_version = jax.__version__
+    required_version = get_minimum_cuda_version_for_jax(jax_version)
+    if cuda_version < required_version:
+        raise RuntimeError(
+            f"Installed CUDA Toolkit: {cuda_version[0]}.{cuda_version[1]} \
+            is not compatible with installed JAX version {jax_version}. \
+            Minimum required CUDA Toolkit for your JAX version \
+            is {required_version[0]}.{required_version[1]}. \
+            Please upgrade your CUDA Toolkit to meet the requirements."
+        )
```
frostedoyster, line 73: I would also mention the possibility of downgrading JAX

frostedoyster: For some users, that will objectively be the most obvious workaround

sphericart-jax/python/sphericart/jax/__init__.py
```diff
@@ -1,18 +1,77 @@
 import jax
+from packaging import version
```
Luthaf, line 2: do we already have a runtime dependency on packaging?

Luthaf: No, we should add it here: https://github.com/lab-cosmo/sphericart/blob/f7b5b710684a04b5895942927aaf79b85d7f53ca/sphericart-jax/pyproject.toml#L5C1-L5C13

PR #179, opened 2025-04-06
Allow FloatX for arbitary X

cortner:
There was a bug in the basis construction, now FloatX for arbitrary X should work, added tests for Float32 and Float128. (also good to test for numerical stability)


cortner:
just merging and will release as 0.1.x

PR #181, opened 2025-04-17
Upgrade tests to Ubuntu 22.04

frostedoyster:
Ubuntu 20.04 is deprecated on GH actions, plus our tests were using a mix of 20.04 and 22.04, now it's all 22.04


Luthaf:
Regarding the test failure: maybe you can pin to an older JAX version for now, so we can update the CI, and then lift the constrain in a separate PR

Luthaf (approved):

PR #182, opened 2025-04-17
Update JAX primitives to use `jax.extend`

frostedoyster:


frostedoyster:
Closing in favor of #188

PR #189, opened 2025-08-07
Update license fields and bump to Py3.9

PicoCentauri:
Otherwise setuptools will prevent installations on newer versions.


ceriottm (approved):
OK 3.8 has been EOL for one year now I guess we can bump our dependency checks up

ceriottm:
Thanks @PicoCentauri for the maintainance

PR #191, opened 2025-08-13
CSCS Compatibility

rubber-duck-debug:
This PR fixes two things:

- when there is a cuda driver mismatch on the system vs. what sphericart has been compiled with. For example when the driver max supported toolkit is 12.4, but code has been compiled with 12.6. Fix here is to write cubin code with NVRTC directly, rather than writing PTX and having NVRTC attempt to jit-compile it to SASS.

- removed the cudnn hack so things work in write-disabled environments. Our wheels support at minimum pytorch 2.4 so I think this is fine.


ceriottm (approved):
LGTM, as long as tests pass and CI runs.

Luthaf:
The code looks fine and the CuDNN hack should no longer be needed if we can build all wheels on CI. If someone complains we can add it back later.

PR #195, opened 2025-08-27
Try to build ARM wheels

frostedoyster:


Luthaf (approved):
The libraries in the wheels look fine, even for CUDA: 

```
nm sphericart/torch/torch-2.4/lib/libsphericart_torch_cuda_stream.so | c++filt | grep " U "
[...]
                 U c10::cuda::getCurrentCUDAStream(signed char)
```

I'll let more people test the wheels manually if they want!

PR #200, opened 2025-10-22
Fix CSCS CI with the new CUDA_HOME changes

Luthaf:
I also made it a required check so we don't forget to run it.


Luthaf:
cscs-ci run

PR #218, opened 2026-03-28
Add CuPy interface

### LINKED ISSUE #217: CuPy support
It should be relatively easy to support CuPy in addition to NumPy. See e.g. https://github.com/lab-cosmo/mops/pull/37

frostedoyster:
As in the title; closes #217. The implementation follows https://github.com/lab-cosmo/mops/pull/37. CuPy support is also made available through the pre-built wheels

### DIFF
--- a/python/src/sphericart/_c_lib.py
+++ b/python/src/sphericart/_c_lib.py
@@ -22,6 +22,22 @@ class sphericart_solid_harmonics_calculator_f_t(ctypes.c_void_p):
     pass
 
 
+class sphericart_cuda_spherical_harmonics_calculator_t(ctypes.c_void_p):
+    pass
+
+
+class sphericart_cuda_spherical_harmonics_calculator_f_t(ctypes.c_void_p):
+    pass
+
+
+class sphericart_cuda_solid_harmonics_calculator_t(ctypes.c_void_p):
+    pass
+
+
+class sphericart_cuda_solid_harmonics_calculator_f_t(ctypes.c_void_p):
+    pass
+
+
 def setup_functions(lib):
     lib.sphericart_spherical_harmonics_new.restype = (
         sphericart_spherical_harmonics_calculator_t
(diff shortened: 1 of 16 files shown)


frostedoyster:
cscs-ci run

frostedoyster:
cscs-ci run

frostedoyster:
cscs-ci run

frostedoyster:
I've downloaded and tested the pre-built linux-x86 wheel and it works

python/src/sphericart/_c_lib.py
```diff
@@ -221,6 +237,168 @@ def setup_functions(lib):
...
+    lib.sphericart_cuda_spherical_harmonics_compute_array_with_hessians.restype = None
+    lib.sphericart_cuda_spherical_harmonics_compute_array_with_hessians.argtypes = [
+        sphericart_cuda_spherical_harmonics_calculator_t,
+        ctypes.POINTER(ctypes.c_double),
+        ctypes.c_size_t,
+        ctypes.POINTER(ctypes.c_double),
+        ctypes.POINTER(ctypes.c_double),
+        ctypes.POINTER(ctypes.c_double),
+        ctypes.c_void_p,
+    ]
```
Luthaf, line 309: This is fine for now, but should we look into simplifying the API a bit? Maybe adding explicit `device` parameter like vesin, and a separate sph_kind="SphericalHarmonics | SolidHarmonics"?

tox.ini
```diff
@@ -37,6 +37,18 @@ deps =
...
+        CUDA_MAJOR=$(nvcc --version | sed -n "s/.*release \([0-9][0-9]*\)\..*/\1/p" | head -n 1); \
+        if [ "$CUDA_MAJOR" = "12" ]; then \
+            python -m pip install -U cupy-cuda12x; \
+        elif [ "$CUDA_MAJOR" = "13" ]; then \
+            python -m pip install -U cupy-cuda13x; \
+        elif [ -n "$CUDA_MAJOR" ]; then \
+            echo "Unsupported CUDA major version for CuPy wheels: $CUDA_MAJOR"; \
+            exit 1; \
+        fi; \
+    fi'
```
Luthaf: If you want, I have a slightly more robust script on vesin, that does not required having the cuda toolkit installed, only the drivers: https://github.com/Luthaf/vesin/blob/main/scripts/install-cupy.sh

Luthaf (commented):
Looks good overall, but we might want to reduce the number of functions we have in the C API; especially if we add a new device through SyCL. It should also make things like #219 which need to update every function easier

python/src/sphericart/_c_lib.py
frostedoyster: This one requires defining a device in the C API, right? Can we do it in a different PR if we need it?

Luthaf: Yes and yes!

frostedoyster:
cscs-ci run

Issue #216, opened 2026-03-28
Torch-compile support


frostedoyster:
The sphericart-torch package currently only supports torchscript, which is deprecated. It should be updated to support `torch.compile` as well.

frostedoyster:
Fixed in #219

PR #219, opened 2026-03-28
Add support for `torch.compile`

frostedoyster:
Addresses #216


sphericart-torch/src/torch.cpp
```diff
@@ -59,63 +112,35 @@ std::vector<torch::Tensor> SolidHarmonics::compute_with_hessians(torch::Tensor x
 }
 
 TORCH_LIBRARY(sphericart_torch, m) {
-    m.class_<SphericalHarmonics>("SphericalHarmonics")
```
Luthaf, line 62: This will be a breaking change for already exported models, are we fine with it?

frostedoyster:
cscs-ci run

sphericart-torch/src/torch.cpp
frostedoyster: I believe that this needs to be done sooner or later because `torch.compile` is much more functional-friendly under the hood, and it's better to break things now rather than later. For now, I think we're safe because no models rely on our `torch.jit.saved` features as far as I know

frostedoyster:
cscs-ci run

ceriottm:
Does this bring speedups?

sphericart-torch/src/torch.cpp
Luthaf: > For now, I think we're safe because no models rely on our torch.jit.saved features as far as I know

Any model that does 

```
class Model(torch.nn.Module):
    def __init__(self):
        self.sph = SphericalHarmonics(...)
```

Would that not include PhACE & anything in torch-spex? Although maybe these were not distributed as torch script exported modules that often …

sphericart-torch/python/sphericart/torch/__init__.py
```diff
@@ -68,5 +69,6 @@ def _lib_path():
         )
 
 
-# load the C++ operators and custom classes
+# load the shared library containing the C++ operators
 torch.classes.load_library(_lib_path())
```
Luthaf: We could move this load_library call inside `register`

sphericart-torch/python/sphericart/torch/register.py
```diff
@@ -0,0 +1,93 @@
...
+def _register_autograd(prefix):
+    gradients = getattr(torch.ops.sphericart_torch, f"{prefix}_with_gradients").default
+    hessians = getattr(torch.ops.sphericart_torch, f"{prefix}_with_hessians").default
+
+    def setup_context(ctx, inputs, output):
+        xyz, ctx.l_max = inputs
+        saved = (xyz,) if isinstance(output, torch.Tensor) else (xyz, output[1])
+        ctx.save_for_backward(*saved)
+
+    def backward(ctx, *grad_outputs):
```
Luthaf, line 91: I guess we loose the ability to call .backward when using the pure C++ API (i.e. not even defining a model in Python). This might be fine for our users

Luthaf (commented):
Looks good overall! 

As a side comment, this is technically a breaking change, and maybe we should release it as sphericart 2.0?

sphericart-torch/src/torch.cpp
frostedoyster: PhACE uses a slow torch implementation for now. Yes for torch-spex but I don't think we've published any models with it (?)

sphericart-torch/python/sphericart/torch/register.py
frostedoyster: Yes but we have to embrace modernity :-)

sphericart-torch/python/sphericart/torch/__init__.py
frostedoyster: It will be done

sphericart-torch/python/sphericart/torch/register.py
frostedoyster: People who want to do libtorch stuff can still use the C++ API with some extra pain

frostedoyster:
cscs-ci run

Luthaf (approved):

PR #221, opened 2026-04-02
Use gpulite

RMeli:


sphericart-jax/src/sphericart_jax_cuda.cpp
```diff
@@ -11,11 +11,10 @@
 #include <memory>
 #include <mutex>
 
-#include "dynamic_cuda.hpp"
+#include <algorithm>
```
RMeli: This is needed because of a missing header in `gpulite`.

Luthaf: Should we update gpulite instead?

RMeli: Yep. I already pushed (by mistake) the [fix to main](https://github.com/rubber-duck-debug/gpu-lite/commit/adf0fa286784e1a469f81b572dd1f1dbd6ef6149), but it ~will need~ would be nice to have a new release.

Luthaf: Sure, I can do a 1.0.4 release! Maybe I'll wait until CI passes on this PR, so we can add all required changes at once!

PR #222, opened 2026-04-15
SYCL follow-up

frostedoyster:
As promised in #210. My impression is that the DTYPE->T change isn't needed anymore after we remove some dead code that was using DTYPE inside the SYCL implementation. Using DTYPE is consistent with all tests, examples and benchmarks elsewhere in sphericart (internally, T is used instead, except for the torch-based implementation which uses torch's "idiomatic" `scalar_t`)


frostedoyster:
@alvarovm The tests look good, let me know if the changes are fine for you (how were these definitions meant to be used initially?)

Luthaf (approved):

alvarovm:
@frostedoyster sorry for the delay. Yes, changes look good to me!

frostedoyster:
cscs-ci run

PR #225, opened 2026-04-26
Build Windows wheels

### LINKED ISSUE #215: Build Windows wheels
After #211, sphericart builds and runs on Windows. The next step would be to provide pre-built wheels for our Python packages, including CUDA support

frostedoyster:
Closes #215


sphericart-jax/setup.py
```diff
@@ -63,8 +50,10 @@ def run(self):
             "-DCMAKE_PLATFORM_NO_VERSIONED_SONAME=ON",
         ]
 
-        CUDA_HOME = _detect_cuda_home()
-        if CUDA_HOME is None:
+        # CUDA support uses gpulite for dynamic loading: no CUDA toolkit needed
+        # at build time. Disable on macOS (no CUDA) and Windows (gpulite's
+        # CUDA DLL discovery is Linux-only, and JAX has no Windows CUDA support).
+        if sys.platform.startswith("darwin") or sys.platform.startswith("win"):
```
Luthaf, line 56: You should be able to compile the code unconditionally, and if someone tries to use it on an unsupported platform, they will get an error about not being able to load libcuda & co.

> CUDA DLL discovery is Linux-only

This seems to be an LLM hallucination, and is not true ^_^ https://github.com/rubber-duck-debug/gpu-lite/pull/9

sphericart-torch/python/sphericart/torch/spherical_hamonics.py
```diff
@@ -67,7 +74,10 @@ def forward(self, xyz: Tensor) -> Tensor:
             spherical harmonics with ``(l, m) = (0, 0), (1, -1), (1, 0), (1,
             1), (2, -2), (2, -1), (2, 0), (2, 1), (2, 2)``, in this order.
         """
-        return torch.ops.sphericart_torch.spherical_harmonics(xyz, self._l_max)
+        stream: int = 0
+        if not torch.jit.is_scripting() and not torch.compiler.is_compiling():
```
Luthaf: This seems wrong, we want the stream to be passed around even in torch script/torch compile mode

sphericart/src/cuda_base.cpp
```diff
@@ -1,14 +1,29 @@
...
+};
+static std::string build_cuda_code() {
+    std::string s;
+    for (const char* const* p = CUDA_PARTS; *p; ++p) {
+        s += *p;
+    }
+    return s;
+}
+static const std::string CUDA_CODE_STR = build_cuda_code();
+static const char* CUDA_CODE = CUDA_CODE_STR.c_str();
```
Luthaf: The way we do this in vesin might be a bit cleaner: 

- https://github.com/Luthaf/vesin/blob/fbcb77d0698a3fb8787de790d955285baac8d2e1/vesin/CMakeLists.txt#L51-L66
- https://github.com/Luthaf/vesin/blob/fbcb77d0698a3fb8787de790d955285baac8d2e1/vesin/cmake/make_hex_array.cmake
- https://github.com/Luthaf/vesin/blob/fbcb77d0698a3fb8787de790d955285baac8d2e1/vesin/src/vesin_cuda.cpp#L32-L34

sphericart-jax/setup.py
frostedoyster: Yes, it's wrong and I was about to remove it

sphericart-torch/python/sphericart/torch/spherical_hamonics.py
frostedoyster: It is passed, but a default value of 0 is chosen during the ghost compile/script run that just traces the shapes of the tensors

sphericart-torch/python/sphericart/torch/spherical_hamonics.py
```diff
@@ -4,6 +4,13 @@
 from torch import Tensor
 
 
+@torch.jit.ignore
+def _current_stream(xyz: Tensor) -> int:
```
RMeli:
```suggestion
def _current_stream(xyz: Tensor):
```

sphericart-torch/python/sphericart/torch/register.py
```diff
@@ -84,30 +84,31 @@ def _register_autograd(prefix):
...
         ctx.save_for_backward(*saved)
 
     def backward(ctx, *grad_outputs):
         if not ctx.needs_input_grad[0]:
-            return None, None
+            return None, None, None  # xyz, l_max, stream
 
         saved = ctx.saved_tensors
         xyz = saved[0]
+        stream = torch.cuda.current_stream(xyz.device).cuda_stream if xyz.is_cuda else 0
```
RMeli: Should we use `None` instead of `0` on the Python side?

sphericart-jax/setup.py
```diff
@@ -63,8 +50,10 @@ def run(self):
             "-DCMAKE_PLATFORM_NO_VERSIONED_SONAME=ON",
         ]
 
-        CUDA_HOME = _detect_cuda_home()
-        if CUDA_HOME is None:
+        # CUDA support uses gpulite for dynamic loading: no CUDA toolkit needed
+        # at build time. Disable on macOS (no CUDA) and Windows (gpulite's
+        # CUDA DLL discovery is Linux-only, and JAX has no Windows CUDA support).
```
RMeli: Is this the case?

sphericart-torch/python/sphericart/torch/spherical_hamonics.py
Luthaf: Then this should only be `torch.compiler.is_compiling()`, without the `is_scripting` check

sphericart-jax/setup.py
frostedoyster: No, it's not the case, my fault. I've changed the comment

sphericart/src/cuda_base.cpp
frostedoyster: Done

sphericart-torch/python/sphericart/torch/register.py
frostedoyster: In principle yes, but we have to feed an int stream to C++, so might as well treat it as an int from the beginning. This value is never used in Python

RMeli: Yes I've seen the C++ side, it just reads a bit odd here. But it's just a minor comment, so happy to discard it.

RMeli:
cscs-ci run

.github/workflows/build-torch-wheels.yml
```diff
@@ -92,6 +96,14 @@ jobs:
...
         python-version: "3.12"
 
+    - name: Install CUDA toolkit (Windows)
+      if: matrix.os == 'windows-2022'
+      uses: Jimver/cuda-toolkit@v0.2.29
+      with:
+        cuda: '12.6.3'
+        method: 'local'
+        log-file-suffix: 'torch-${{ matrix.pytorch-version }}-${{ matrix.os }}.txt'
+
```
Luthaf, line 105: This is required to build `sphericart_cuda_stream.dll`, right?

.github/workflows/build-torch-wheels.yml
```diff
@@ -101,7 +113,7 @@ jobs:
         CIBW_BUILD_VERBOSITY: 1
         CIBW_BUILD: "cp312-*"
         CIBW_SKIP: "*-musllinux*"
-        CIBW_ARCHS: "${{ matrix.arch }}"
+        CIBW_ARCHS: "${{ matrix.cibw_arch || matrix.arch }}"
```
Luthaf: Can we just add `cibw_arch` to all matrix entries for clarity?

setup.py
```diff
@@ -59,8 +46,9 @@ def run(self):
             "-DCMAKE_PLATFORM_NO_VERSIONED_SONAME=ON",
         ]
 
-        CUDA_HOME = _detect_cuda_home()
-        if CUDA_HOME is None:
+        # CUDA support uses gpulite for dynamic loading: no CUDA toolkit needed
+        # at build time. Disable only on macOS where CUDA is unsupported.
+        if sys.platform.startswith("darwin"):
             cmake_options.append("-DSPHERICART_ENABLE_CUDA=OFF")
```
Luthaf, line 52: Fine for now, but we might want to remove `SPHERICART_ENABLE_CUDA` altogether, and just always compile the code. It would then fail at runtime on platforms without cuda, but this is fine since they don't have cuda =)

Luthaf (approved):
Looks good, just a small suggestion to cleanup the github action yaml

.github/workflows/build-torch-wheels.yml
frostedoyster: Yes, exactly

setup.py
frostedoyster: I guess we ship less useless binary this way, but I will review a further PR if you want to remove it

PR #247, opened 2026-06-23
Do not bundle nvrtc in wheels

Luthaf:


pfebrer (approved):
This will fix metatrain ci

PR #252, opened 2026-07-09
build wheels for torch 2.13

sofiia-chorna:


Luthaf:
The CI failure is the expected one with C++20

Luthaf:
Thanks @sofiia-chorna!

